PR #5, opened 2015-09-04
Improve table cell text visibility

Poorchop:
-Use dark text color against light background

untitled

### DIFF
--- a/wikipedia-dark.css
+++ b/wikipedia-dark.css
@@ -167,6 +167,10 @@
     color: #ccc !important;
   }
 
+  .table-no, .table-yes, .table-partial {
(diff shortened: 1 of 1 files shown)


Mottie:
Thanks!

Mottie:
Actually, I believe `.table-yes` and `.table-no` are new definitions... so it might be better to make the background color darker instead of the text lighter. Can you share a link to that page. I can't seem to quickly find `.table-partial` in the examples I've seen so far.

Poorchop:
Guessing you found an example already but here's [the page](https://en.wikipedia.org/wiki/Office365) that I was referencing.

PR #7, opened 2015-09-05
Add dark background to even rows of infobox

Poorchop:
[Example](https://en.wikipedia.org/wiki/EarthBound)

The default CSS uses alternating colors between the even and odd rows so I did the same thing here.


Mottie:
Thanks... you know you have privileges you could merge these yourself.

Mottie:
Oops, you have privileges now :P

PR #9, opened 2015-09-08
Dark background and brighter text for table footer

Poorchop:
@Mottie wanted to run this by you first to see if you agree with this change and these selectors.

[Link](https://en.wikipedia.org/wiki/Ted_Cruz#Electoral_history)


Mottie:
The changes look good to me :+1:

Mottie:
I have no idea why we haven't merged this PR until now... LOL

Poorchop:
It got lost in purgatory.

PR #16, opened 2016-02-08
Added .mwe-popups, .mwe-popups-is-not-tall, .mwe-popups-is-tall

chj1axr0:
to the background-color: #222; list


Mottie:
Hi @chj1axr0!

Would you share a link to the page where I can find the popup with those classes? Thanks!

chj1axr0:
@Mottie The popups are part of the [Beta](https://en.wikipedia.org/wiki/Special:Preferences#mw-prefsection-betafeatures) section under hovercards

Mottie:
Thanks!

PR #19, opened 2016-12-08
added Related pages from beta

chj1axr0:
Added the [Related pages](https://en.wikipedia.org/wiki/Special:Preferences#mw-prefsection-betafeatures) that show up on the bottom of the page to the appropriate background sections


Mottie:
Thanks!

PR #29, opened 2018-01-01
Invert/Darken logo images and update README

Aced14:
These changes completely invert the logo images' colours, then darken the result to make the whitest parts of the inverted images identical to the theme's light grey font colour (#9a9a9a).

References in the README about it being impractical to adjust the logo images were also removed.

**Screenshots...**

**Main index page - before:**
index-before

**Main index page - after:**
index-after

**English index page - before:**
index-en-before

**English index page - after:**
index-en-after


Mottie:
Hi @Aced14!

I'm not a big fan of the inverted logo. I just pushed an update that inverts the logo by 33%, which isn't great, but the text is more readable.

Aced14:
@Mottie So your personal preference would be to keep the globe completely unaltered, but make the text in those images brighter (ideally the same colour as the standard font colour)?

If so, I could send over another PR that'd be able to do that. I was able to come up with a way to use pseudo-elements to "split" the top-left logo image into two parts so the globe and text can be manipulated separately. Since that image's dimensions seem to be the same across all languages and the globe's size never changes, it'd be able to work in any language's variant of that image.

Mottie:
Sure! Hmm, maybe we can overlay the globe instead of splitting the image.

Aced14:
@Mottie I'll send over a PR later that shows off exactly what I had in mind :).

PR #31, opened 2018-03-19
Invert color of black loudspeaker icon

Poorchop:
This is for the icons next to the pronunciation links. I just created this rule so I think some testing is in order, but so far I haven't noticed it inverting any icons that should be left alone.

Here is the [example page](https://en.wikipedia.org/wiki/Wikipedia) - you should see the speaker icons next to where it explains how to pronounce Wikipedia.


Mottie:
Hi @Poorchop!

I'm on it!

PR #35, opened 2018-06-18
fix tooltip/reversed

the-j0k3r:
tooltip

I was trying to find this one 

capture

But cant grab with inspect and css from style editor was nothing I could see useful.

So ended up fixing the others :( that I could find, sadly had to add as is else theres arrows flying outside the box, anyway Ive lost my plot with these tooltips. Argh!

### DIFF
--- a/wikipedia-dark.css
+++ b/wikipedia-dark.css
@@ -247,14 +247,19 @@
   .mwe-popups.flipped-y:before, .mwe-popups.flipped-x-y:before {
     border-top-color: #555 !important;
   }
-  .mwe-popups:after {
+  .mwe-popups:after,
+  .oo-ui-popupWidget-anchored-top .oo-ui-popupWidget-anchor:before,
+  .oo-ui-popupWidget-anchored-top .oo-ui-popupWidget-anchor:after {
     border-bottom-color: #222 !important;
   }
   .mwe-popups.flipped_y:after, .mwe-popups.flipped_x_y:after,
-  .mwe-popups.flipped-y:after, .mwe-popups.flipped-x-y:after {
+  .mwe-popups.flipped-y:after, .mwe-popups.flipped-x-y:after,
+  .oo-ui-popupWidget-anchored-bottom .oo-ui-popupWidget-anchor:before,
+  .oo-ui-popupWidget-anchored-bottom .oo-ui-popupWidget-anchor:after {
     border-top-color: #222 !important;
   }
-  .mwe-popups, .mwe-popups-is-not-tall, .mwe-popups-is-tall {
+  .mwe-popups, .mwe-popups-is-not-tall, .mwe-popups-is-tall,
+  .oo-ui-popupWidget-popup {
     border: 1px solid #555 !important;
   }
   .mwe-popups-settings-icon:hover, .mwe-popups-settings-icon:active {
--- a/wikipedia-dark.css
+++ b/wikipedia-dark.css
@@ -281,20 +281,6 @@
   .referencetooltip > li + li {
     border-top: 12px #555 solid !important;
   }
- 
-  .referencetooltip.RTflipped > li + li {
-    position: absolute !important;
-    top: 8px !important;
-    border-top: 0 !important;
-    border-bottom: 8px #222 solid !important;
-  }
-
-  .referencetooltip.RTflipped > li + li::after {
-    border-top: 0 !important;
-    border-bottom: 8px #222 solid !important;
-    position: absolute !important;
-    margin-top: -3px !important;
-  }
 
   /* Bookmark Icon */
   #ca-watch.icon a {
--- a/wikipedia-dark.css
+++ b/wikipedia-dark.css
@@ -258,12 +258,36 @@
     border-bottom: #555 1px solid !important;
   }
 
+  .referencetooltip > li {
+    background: #222 !important;
+    border: 1px solid #555 !important;
+  }
+
   /* arrow down */
   .referencetooltip > li + li, .referencetooltip > li + li::after {
-    border-top-color: #555 !important;
+    border-top-color: #222 !important;
     border-right-color: transparent !important;
     border-left-color: transparent !important;
   }
+
+  .referencetooltip > li + li {
+    border-top: 12px #555 solid !important;
+  }
+ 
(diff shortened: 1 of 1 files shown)


Mottie:
Hey @the-j0k3r !

Wikipedia has at least two different tooltips ([prefs](https://en.wikipedia.org/wiki/Special:Preferences):

* Navigation popups - these tooltips/popups don't include arrows.
* Reference Tooltips - these tooltips do have arrows, but don't work if navigation popups are enabled.

I'm not seeing the "RTflipped" class at all, even though it seems to have initials for the reference tooltip.

Anyway, we shouldn't need to modify the non-color styles of the tooltip. I tested the following additions on  [this random page](https://en.wikipedia.org/wiki/List_of_University_of_Chicago_Booth_School_of_Business_faculty) and it fixes the downward pointing arrows. You can get the tooltip to flip by hovering over the link *while* it is near the bottom of the page.

```css
.mwe-popups.flipped-y:after {
  border-color: #222 transparent transparent transparent !important;
}
.mwe-popups.flipped-y:before {
  border-color: #555 transparent transparent transparent !important;
}
```

the-j0k3r:
Thanks, the rt flipped is present, its for the reference links like `[2] `

Ill try your fix, thanks.

Also any particular reason why you kill sintax highlighting with `  span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]),` set to `    color: #9a9a9a;`

the-j0k3r:
Also

```
.mwe-popups.flipped-y:after {
  border-color: #222 transparent transparent transparent !important;
}
.mwe-popups.flipped-y:before {
  border-color: #555 transparent transparent transparent !important;
}

```

doesnt fix this

capture

Mottie:
LOL I never said it was perfect! It looks like there is also a `.flipped-x-y` class name. Also, it'd be better to update [lines 245-254](https://github.com/StylishThemes/Wikipedia-Dark/blob/master/wikipedia-dark.css#L245-L254):

```css
  .mwe-popups.flipped_y:before, .mwe-popups.flipped_x_y:before,
  .wikiEditor-toolbar-dialog .ui-dialog-buttonpane, .mwe-popups.flipped-y:before,
  .mwe-popups.flipped-x-y:before {
    border-top-color: #555 !important;
  }
  .mwe-popups:after {
    border-bottom-color: #222 !important;
  }
  .mwe-popups.flipped_y:after, .mwe-popups.flipped_x_y:after, .mwe-popups.flipped-y:after,
  .mwe-popups.flipped-x-y:after {
    border-top-color: #222 !important;
  }
```



Where are you seeing syntax highlighting issues? Share a link please.

the-j0k3r:
There is no syntax highlighting its all whitish cause of that selector I posted.

https://en.wikipedia.org/wiki/Indentation_style

capture

remove `span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]),` and it works.

the-j0k3r:
get rid of the white hover on mwe-popup too with db953ef and tidy selectors.

Mottie:
* I'm still not seeing the `.RTflipped` class... which gadget adds it?
* Please remove the non-color related entries that are already defined, e.g. `position: absolute`, etc.
* Please only modify the `border-color` as needed. You shouldn't need to modify the `solid 1px`, etc.

the-j0k3r:
Some of those entries are needed, but OK Ill remove the lot cause without looks like ass.

the RT flipped I already said its for the reference hover toolitips.

the-j0k3r:
removed what you ask, but the flipped reverted reference tooltip is broken now.as shown in screenshot above.

the-j0k3r:
The best solution without positioning items ouside default values (what I done to fix the issues) is this.

```.css
  .referencetooltip.RTflipped > li + li {
    border-bottom-color: #222 !important;
  }

  .referencetooltip.RTflipped > li + li::after {
    border-bottom-color:  transparent !important;
  }
```
but you dont have any borders on the arrows which I suppose its OK for most ppl.

the-j0k3r:
:fire:  More broken tooltip arrows.

https://en.wikipedia.org/wiki/Special:RecentChangesLinked?hidebots=1&hidecategorization=1&hideWikibase=1&target=Wikipedia%3AFile+Upload+Wizard&limit=50&days=7&urlversion=2

tooltips

and I see the syntax highlighting line thats breaking it belongs to these buttons :)

the-j0k3r:
OK new tooltips fixed... sigh. (its easier to re-write the whole theme, I may still do my own.

Right, syntax highlighting still broken as with many many things still, so this is ready unless I need to do something?

Mottie:
I haven't had a chance to look at the updates yet... 

Maybe to fix syntax highlighting, we can add a paragraph tag in front of the spans? Let me know if this works please:

```css
p span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]), ...
```

the-j0k3r:
Yes that works :)

the-j0k3r:
He no, that fixes syntax like I said, but breaks the buttons it also uses it.

capture

wikipedia-dark.css
```diff
@@ -21,7 +21,7 @@
 
   /*** Overall ***/
   td, p, li, th:not([style*="background"]), caption, div,
-  span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]),
+  p span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]),
   p code, label code, dl code, .oo-ui-widget div, .oo-ui-widget label {
```
Mottie: Wow, my `p` is all grown up.

Mottie:
And ugh, I wish wikipedia would create a style guide and stick with it. The styles and design are all over the place.

Mottie:
Lets just revert the `p` addition &ndash; poor `p` &ndash; Merge and deal with the syntax highlighting another day.

the-j0k3r:
there, Ive had enough of wikipedia too, Im making my own style to support all of them.

the-j0k3r:
> And ugh, I wish wikipedia would create a style guide and stick with it. The styles and design are all over the place.

naaa, thats tech from a million years ago + new tech + vavavoom

Mottie:
Thanks for your hard work!

the-j0k3r:
Youre welcome, :)

PR #46, opened 2018-09-28
Extending the dark theme to cover the scrollbar

HerrHulaHoop:
I was trying to add dark theme for the scrollbar when I noticed your GitHub dark theme already has a neat version of it. So I've decided to use that instead and only removed GitHub specific portions. As with that theme, this addition only works on Chrome.

<details>
 <summary>Before</summary>

before
</details>
<details>
 <summary>After</summary>

after
</details>

Do let me know if anything else is required!


Mottie:
Thanks!

PR #48, opened 2018-10-07
Add background color for user message pop-ups

HerrHulaHoop:
Message notifications that show up as a banner in the wiki page does not have a background making it pretty easy to miss. Added a background to let it stand out.

Original white theme
white

Before
before

After
after

Let me know if you have any comments!

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -187,7 +187,7 @@
   td[style*="background:#F2F2F2" i],
   table.wikitable > tbody > tr[style*="background-color:#F6F6F6" i], .tlheader,
   th[style*="background:whitesmoke" i], td[style*="background:whitesmoke" i],
-  th[style*="#FFEBAD" i] {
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @HerrHulaHoop!

I don't think green is the right color for those popups. Either pick brown, a darker version of the original yellow, or grey.

HerrHulaHoop:
Changed it to grey. Opted for grey as brown didn't stand out enough and yellow made it difficult to read the text.

image

Let me know if this is okay!

HerrHulaHoop:
`#333` seems like a better fit as `#282828` looks is a bit too dark. The notification has to have a little contrast to the regular colors as it's easy to miss it otherwise. I'll go ahead and change it to `#333`.

Mottie:
Ok, add it to an already defined `#333` entry please. There are two, make sure to add it to the one with the `!important` flag.

HerrHulaHoop:
Done 👍

Mottie:
Thanks!

PR #56, opened 2019-03-10
Remove dedicated scrollbar code and add Overlay-Scrollbars to README

HerrHulaHoop:
Since a [dedicated scrollbar theme](https://github.com/StylishThemes/Overlay-Scrollbars) has been made, there is no longer a need to have it as part of this theme. 

The new style allows users to customise the scrollbar from the extension itself and works on Firefox as well. So I think it is a better solution to point users to it rather than baking it into this theme.


Mottie:
Good point! Thanks!

PR #57, opened 2019-04-08
Fix reference tooltip colours

Aced14:
**Screenshots...**

**Before:**
wikipedia-dark-reference-tooltip-before

**After:**
wikipedia-dark-reference-tooltip-after


the-j0k3r:
Thank you for this PR. =)

PR #61, opened 2019-04-20
Improve button styles

Aced14:
* Fix blue/cancel buttons that were showing up as generic ones
* Fix border colours on blue buttons
* Fix hover, active, focus and active+focus styles

**Screenshots...**

**Main page search...**
**Before + After:**
main-page-search-before

**Reference tooltips settings...**
**Before:**
reference-tooltips-config-popup-before
**After:**
reference-tooltips-config-popup-after

**Page preview settings...**
**Before:**
page-preview-config-overlay-before
**After:**
page-preview-config-overlay-after

**Edit page...**

**Normal...**
**Before:**
edit-page-before
**After:**
edit-page-after

**``:hover``**
**Before:**
edit-page-hover-before
**After:**
edit-page-hover-after

**``:active``**
**Before:**
edit-page-active-before
**After:**
edit-page-actve-after

**``:focus``**
**Before:**
edit-page-focus,active-focus-before
**After:**
edit-page-focus-after

**``:active:focus``**
**Before:**
edit-page-focus,active-focus-before
**After:**
edit-page-actve-after

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -68,12 +68,13 @@ domain("wiki.rpcs3.net") {
   }
   /*** Overall ***/
   td, p, li, th:not([style*="background"]), caption, div,
-  span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]),
+  span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]):not([class*="oo-ui-labelElement-label"]),
   p code, label code, dl code, .oo-ui-widget div, .oo-ui-widget label {
     color: #9a9a9a;
   }
   body, h1, h2, h3, h4, h5, h6, #footer li, .infobox th[style*="background"],
   div[style*="color"], input[type="search"], input[type="submit"],
+  .oo-ui-buttonElement-button, .oo-ui-buttonElement-button:hover,
   input[type="number"], select, .suggestions-result, input,
   .central-featured-lang .link-box em, .central-featured-lang .link-box small {
     color: #9a9a9a !important;
@@ -87,8 +88,9 @@ domain("wiki.rpcs3.net") {
   legend .oo-ui-labelElement-label, .cnotice-message p {
     color: #ddd !important;
   }
-  a:hover, a:hover span, #toc a:hover, #toc a:hover span,
-  .suggestions-result .highlight, .suggestions-special .special-query {
+  a:hover, a:hover span:not(.oo-ui-labelElement-label), #toc a:hover,
+  #toc a:hover span, .suggestions-result .highlight,
+  .suggestions-special .special-query {
     color: #fff !important;
   }
   /* text selection */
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Hello @Aced14! Thank you for reaching out with this contribution.

I like the changes proposed, as for the dark text within buttons and icons, all our styles sort have this styling in common, so It may look out of place.

Lets see what others say.. @Mottie @xt0rted @silverwind

Mottie:
Our goal is to keep the dark style as close to the original as possible, but darker. So already blue buttons should be darker blue, bright white text inside those buttons should remain white, just not as bright. The Wikipedia style makes that incredibly hard in that most custom colors are added in-line with no specification to follow.

I don't think we really have hard-and-fast rules, so if you think something looks better one way or another, I say go for it... we can agree or disagree on a case-by-case basis... I'm okay with what you have in this PR so far, so that's an okay vote from me.

Aced14:
Regarding the dark text in blue buttons... my line of thinking was that since Wikipedia's default font colour is really dark grey and is changed to light grey in this dark theme, it's more or less being inverted. By its very nature, the dark theme has to invert colours to some degree to change the site's overall colour scheme from bright to dark - but without going so far as to arbitrarily invert all colours.

In Wikipedia's normal theme, blue buttons seem to coloured as such to indicate a primary action. So to stand out compared to other buttons, they're coloured in bright blue and use an inverted font colour (white).

As far as this PR goes, I felt like using a black font for blue buttons would stay "truer" to what the normal Wikipedia theme was trying to achieve with its button colours (i.e. using an inverted colour to contrast with the default font colour).

the-j0k3r:
My 2cents, Ide say staying true to original is important, but staying visible is more important, too dark  IMO.

Aced14:
@the-j0k3r I'll change it back to white at some point tomorrow.

wikipedia-dark.user.css
```diff
@@ -113,9 +113,32 @@ domain("wiki.rpcs3.net") {
...
   }
-  .mw-ui-button, .mw-ui-button *, button[type="submit"] {
+  .mw-ui-button, .mw-ui-button *, button[type="submit"],
+  .pure-button-primary-progressive,
+  .oo-ui-buttonElement-framed.oo-ui-widget-enabled.oo-ui-flaggedElement-primary.oo-ui-flaggedElement-progressive > .oo-ui-buttonElement-button,
+  .mw-ui-button.mw-ui-progressive {
     background-color: var(--base-color) !important;
-    color: #fff !important;
+    border-color: transparent !important;
+    color: #222 !important;
```
the-j0k3r:
```suggestion
    color: #fff !important;
```

Aced14:
Sorry, will take another day or two :(.

I've revised my commit (haven't pushed to GitHub yet), but want to rewrite its commit message, update the OP's screenshots and add comparison shots of hover/active/focus states. Ran out of time to do it tonight.

the-j0k3r:
> I've revised my commit (haven't pushed to GitHub yet), but want to rewrite its commit message, update the OP's screenshots and add comparison shots of hover/active/focus states. Ran out of time to do it tonight.

No problem, if you want to do all of that, CSS change is all we need.

Aced14:
Just forced-pushed an updated commit and reworked the OP.

The updated commit contains more changes than just the white text. So it might be worth re-reviewing in its entirety. I eliminated brightness filters since they were causing noticeable text blurring. Also tried moving things around to better fit within the different comment sections.

the-j0k3r:
Thank you for your contribution. =)

PR #62, opened 2019-04-20
Support page preview settings

Aced14:
**Screenshots...**

**Settings...**

**Default:**
page-preview-default

**Before:**
page-preview-before

**After:**
page-preview-after-v2

**Disable notice...**

**Default:**
page-preview-close-default

**Before:**
page-preview-close-before

**After:**
page-preview-close-after

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -70,7 +70,7 @@ domain("wiki.rpcs3.net") {
   td, p, li, th:not([style*="background"]), caption, div,
   span:not([class*="color"]):not([class*="wikEd"]):not([style*="background"]):not([class*="oo-ui-labelElement-label"]),
   p code, label code, dl code, .oo-ui-widget div, .oo-ui-widget label,
-  #pagehistory li.selected {
+  #pagehistory li.selected, #mwe-popups-settings main form label {
     color: #9a9a9a;
   }
   body, h1, h2, h3, h4, h5, h6, #footer li, .infobox th[style*="background"],
@@ -86,7 +86,9 @@ domain("wiki.rpcs3.net") {
   a:visited, #toc a:visited, #toc a:visited span {
     color: var(--visited-color) !important;
   }
-  legend .oo-ui-labelElement-label, .cnotice-message p {
+  legend .oo-ui-labelElement-label, .cnotice-message p,
+  #mwe-popups-settings header h1, #mwe-popups-settings main form label > span,
+  .mwe-popups-settings-help p {
     color: #ddd !important;
   }
   a:hover, a:hover span:not(.oo-ui-labelElement-label), #toc a:hover,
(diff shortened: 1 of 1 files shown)


Aced14:
Just force-pushed an update with a few corrections:
* Moved new selectors the end of pre-existing selector groups
* Ensured modified lines don't exceed 80 characters
* Replaced a random tab character with 4 spaces

Aced14:
@the-j0k3r @Mottie Any thoughts on **Page preview/Enable/Disable**'s font colour?

In Wikipedia's default theme, it's ``#000000``. It's currently ``#9a9a9a`` in this PR (i.e. this theme's normal font colour). I could change it to ``#ffffff`` if you'd prefer.

wikipedia-dark.user.css
```diff
@@ -447,7 +450,8 @@ domain("wiki.rpcs3.net") {
             filter: invert(100%) !important;
     background-color: transparent !important;
   }
-  .central-textlogo__image, .unicode.haudio a img {
+  .central-textlogo__image, .unicode.haudio a img,
+  .mw-ui-icon-popups-close::before {
```
the-j0k3r: Single colon as we do with other similar entries. =)

wikipedia-dark.user.css
```diff
@@ -463,6 +467,13 @@ domain("wiki.rpcs3.net") {
   .rt-tooltipTail {
     background: #555 !important;
   }
+  /*** Page preview settings ***/
+  .mwe-popups-overlay {
+    background-color: rgba(34, 34, 34, 0.75) !important;
+  }
+  #mwe-popups-settings {
+    box-shadow: 0 2px 2px 0 rgba(85, 85, 85, 0.25) !important;
```
the-j0k3r: We dont prefix with the zero.

wikipedia-dark.user.css
Aced14: Wasn't sure since it seemed like a mix of both one and two colon selectors were already in use for pseudo-elements. ``::selection`` and ``::-moz-selection`` are using two.

Single-colon syntax for pseudo-elements is also [deprecated in CSS3](https://www.w3.org/TR/selectors/#pseudo-element-syntax).

wikipedia-dark.user.css
Aced14: Fixed :).

Aced14:
@the-j0k3r I've updated my commit to address your line comments and removed the reference to #61 from the OP (since it's no longer relevant).

Haven't had a chance to look into brightening-up the text yet. Will try to look into it within the coming days.

Aced14:
@the-j0k3r Updated again and revised the OP's after screenshot. Went with ``#ddd`` in the end.

Also applied it to a paragraph that appears in the popup if the user disables the setting and saves. Surprisingly, that paragraph's colour wasn't declared at all in Wikipedia's normal CSS and wasn't inherited from anything.

Aced14:
One more force-push... Partially-inverted the disable notice's sample image to blend in better with its surroundings. Also added comparison screenshots of it to the OP.

the-j0k3r:
Hello again ;), this turned into a wack-a-mole game, this tends to happen when multiple PR's happen to same source file. To avoid the wack-a-mole rebase and fix game in future, try to group fixes in same PR but under different commits correspondingly named.

Now for this, If you could rebase and solve the conflicts we can get this merged asap and do a release =)

Thanks in advance.

Aced14:
Just rebased this PR :).

Regarding grouping commits in the same PR, I usually try to separate my PRs so that each one is dedicated to a specific kind of change. That way, I can focus entirely on what's needed for one kind of change and easily force-push replacement commits as needed.

In any case, I don't think my future PRs will be as likely to run into conflicts as the ones I recently sent in. So hopefully things will be easier to deal with next time around.

the-j0k3r:
Thanks for your contribution, again =)

PR #64, opened 2019-04-27
Minor fixes

Atexor:
infobox not themed properly on polish version in Wikipedia, unecessary changing img color when focused, added color to "Not logged in"


the-j0k3r:
Thank you for reaching out to us with this contribution! =)

PR #66, opened 2019-05-14
Fix quotes in Russian articles

TotalCaesar659:
I've found that many articles in Russian (I didn't see this in English ones) don't apply dark style to their quotes. For example: Martin Niemöller's article (https://ru.wikipedia.org/wiki/Нимёллер,_Мартин)
Before:

After:

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -428,7 +428,7 @@ domain("wiki.rpcs3.net") {
     color: #ff4242 !important;
   }
   /*** Quotes ***/
-  .ts-Начало_цитаты-quote {
+  blockquote {
     background-color: transparent !important;
     border: none !important;
   }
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -427,6 +427,11 @@ domain("wiki.rpcs3.net") {
   .oo-ui-buttonElement-frameless.oo-ui-widget-enabled.oo-ui-flaggedElement-destructive > .oo-ui-buttonElement-button:hover {
     color: #ff4242 !important;
   }
+  /*** Quotes ***/
+  .ts-Начало_цитаты-quote {
```
Mottie: Hi @TotalCaesar659!

Thanks for reporting this issue and for your contribution!

It looks like a more generic `blockquote` can be used instead of `.ts-Начало_цитаты-quote` to fix that style. Please update this PR and I'll merge it. Thanks!

TotalCaesar659: Sorry, I had a doubt about it, I'm not so good in changing HTML styles. Fixed.

PR #67, opened 2019-05-15
Fix review notice

TotalCaesar659:
Text of review notices is almost invisible, e.g. https://ru.wikipedia.org/wiki/Чайф ("The current article version hasn't been reviewed yet by skilled contributors and may differ from one checked 12 March 2019; 1 edit has to be checked.").
Before

After

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -78,7 +78,7 @@ domain("wiki.rpcs3.net") {
   div[style*="color"], input[type="search"], input[type="submit"],
   .oo-ui-buttonElement-button, .oo-ui-buttonElement-button:hover,
   input[type="number"], select, .suggestions-result, input,
-  .central-featured-lang .link-box em, .central-featured-lang .link-box small {
+  .central-featured-lang .link-box em, .central-featured-lang .link-box small, div.flaggedrevs_preview {
     color: #9a9a9a !important;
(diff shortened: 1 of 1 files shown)


Mottie:
Hi @TotalCaesar659!

Thanks for your contribution! Please move the selector up to [this block](https://github.com/StylishThemes/Wikipedia-Dark/blob/master/wikipedia-dark.user.css#L77-L83) (to the end of line 81). Our convention is to combine common definitions and to add all new selectors to end of the list. We may not always follow this rule strictly, but it keep from having way too many repeated definitions.

Also note that there are two different definitions for `color: #9a9a9a`; one with and one without an `!important` flag.

TotalCaesar659:
Fixed, and what's the difference between them?

the-j0k3r:
> Fixed, and what's the difference between them?

The difference is instead of having two rules with same attributes we only have one. Basically what @mottie said.

Mottie:
I think they were asking why there is a color definition with and without an important flag. The one without an important flag is applied to general elements, like all table cells; this allows the color of the cell to fit our dark style and not override any inline style color settings on table cells that have a background color.

the-j0k3r:
Thank you for this contribution =) I manually added your fix due to a conflict in PR, it was faster this way.

PR #69, opened 2019-05-16
Add wiki.archlinux.org

polyzen:


the-j0k3r:
Hi @polyzen, thank you for reaching out to us with this contribution.

I added this temporarily to have a quick test, it seems though theres a great deal more behind Archlinux wiki site that has been customized, link colors and other elements and simply adding the domain here is not going to be enough.

That said, and because I dont actually use this theme, maybe @Mottie can decide whats the best thing regarding this.

Mottie:
Hi @polyzen!

Thanks for your contribution! I think @the-j0k3r is right, we could add the domain, but there is a bit more work to do to have the dark style's customization work on the archlinux site. I'll take a look this weekend.

PR #70, opened 2019-05-22
Properly style positive value in Recent Changes page

HerrHulaHoop:
Before:
Before

After:
After


the-j0k3r:
Thank you for this contribution.

PR #71, opened 2019-05-23
Better list of wikimedia sites

AfroThundr3007730:
Adds all the other Wikimedia projects that were missing, plus explicitly specify the *.wikimedia.org wikis without accidentally applying to non-wiki sites (e.g. ticket.wikimedia.org).

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -39,8 +39,18 @@
 @-moz-document domain("wikipedia.org"),
 domain("wikidata.org"),
 domain("mediawiki.org"),
-domain("wikimedia.org"),
+domain("wikitech.wikimedia.org"),
+domain("otrs-wiki.wikimedia.org"),
+domain("commons.wikimedia.org"),
+domain("meta.wikimedia.org"),
+domain("species.wikimedia.org"),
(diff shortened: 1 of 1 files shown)


AfroThundr3007730:
We could probably just put that into a `regexp()` instead of multiple `domain()` clauses, so something like:

    regexp(https?:\/\/[^?]*((wikipedia|mediawiki|wikibooks|wikidata|wikinews|wikiquote|wikisource|wikiversity|wikivoyage|wiktionary)|(meta|commons|otrs-wiki|species|wikitech)\.wikimedia)\.org\/.+)

Although that's way less legible than the domain list.

the-j0k3r:
@AfroThundr3007730 Thanks for this. I think the regex would be nicer, but lets see what @Mottie says.

Mottie:
HI @AfroThundr3007730!

Thanks for helping out! I'm fine with either way; but no need to give yourself extra work. We can merge this PR as is.

the-j0k3r:
Also, I presume that you have checked these extra domains are on par coverage wise, if they have customized any classes this will end up being a bit like the Archlinux site addition on the other PR.

PR #73, opened 2019-05-27
Minor style tweaks

AfroThundr3007730:
- Fixed notification UI "mark as read/unread" button colors
  - Fixed background and padding of "More" menu elements
  - Reduced brightness of the footer icon images
  - Fixes background colors on the popups gadget
  - Uses a regex for the domain selector instead of a domain list

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -37,21 +37,21 @@
 @preprocessor uso
 ==/UserStyle== */
 @-moz-document domain("wikipedia.org"),
-domain("wikidata.org"),
-domain("mediawiki.org"),
-domain("wikitech.wikimedia.org"),
-domain("otrs-wiki.wikimedia.org"),
-domain("commons.wikimedia.org"),
-domain("meta.wikimedia.org"),
-domain("species.wikimedia.org"),
-domain("wiktionary.org"),
-domain("wikivoyage.org"),
-domain("wikiversity.org"),
-domain("wikinews.org"),
-domain("wikiquote.org"),
-domain("wikisource.org"),
-domain("wikibooks.org"),
-domain("wiki.rpcs3.net") {
+               domain("wikidata.org"),
+               domain("mediawiki.org"),
+               domain("wikitech.wikimedia.org"),
+               domain("otrs-wiki.wikimedia.org"),
+               domain("commons.wikimedia.org"),
+               domain("meta.wikimedia.org"),
+               domain("species.wikimedia.org"),
+               domain("wiktionary.org"),
+               domain("wikivoyage.org"),
+               domain("wikiversity.org"),
+               domain("wikinews.org"),
+               domain("wikiquote.org"),
+               domain("wikisource.org"),
+               domain("wikibooks.org"),
+               domain("wiki.rpcs3.net") {
   /* Modified from https://userstyles.org/styles/47161/dark-wikipedia-rounded */
   :root {
     --base-color: /*[[base-color]]*/;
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -37,21 +37,21 @@
...
+               domain("meta.wikimedia.org"),
+               domain("species.wikimedia.org"),
+               domain("wiktionary.org"),
+               domain("wikivoyage.org"),
+               domain("wikiversity.org"),
+               domain("wikinews.org"),
+               domain("wikiquote.org"),
+               domain("wikisource.org"),
+               domain("wikibooks.org"),
+               domain("wiki.rpcs3.net") {
```
the-j0k3r: This maybe unnecessary adding so many white spaces here See [Style Guide](https://github.com/StylishThemes/Wikipedia-Dark/blob/master/CONTRIBUTING.md#style-guide) in contributing.md, This may throw perfectionist into a fit, idk will test it out later)

the-j0k3r: Actually perfectionist didn't throw a fit, I think however this would look more in the styleguide as this below.
```css
@-moz-document
  domain("wikipedia.org"),
  domain("wikidata.org"),
  domain("mediawiki.org"),
  domain("wikitech.wikimedia.org"),
  domain("otrs-wiki.wikimedia.org"),
  domain("commons.wikimedia.org"),
  domain("meta.wikimedia.org"),
  domain("species.wikimedia.org"),
  domain("wiktionary.org"),
  domain("wikivoyage.org"),
  domain("wikiversity.org"),
  domain("wikinews.org"),
  domain("wikiquote.org"),
  domain("wikisource.org"),
  domain("wikibooks.org"),
  domain("wiki.rpcs3.net") {
```

AfroThundr3007730: @the-j0k3r That would probably be better, yes, but perfectionist was the one that put them there.
It seems to be adhering to the practice of lining up arguments to a function for readability:
```python
def my_function_name(arg1=None, arg2='Foo', bar=Baz,
                     yet_another_arg='Something else',
                     one_more_arg=42)
```
I'm not sure how to override it. :/

AfroThundr3007730:
Too much breakage with my first method, second commit brings it all together correctly.

wikipedia-dark.user.css
AfroThundr3007730: Although, I suppose the aforementioned `regexp()` syntax would obviate the need to fix this.

the-j0k3r: Perfectioninst wont affect this part, though for readability what I suggested feels more inline with the style guide and overall indenting.

the-j0k3r:
> Too much breakage with my first method, second commit brings it all together correctly. Learning a lot more about CSS and these selector specificity shenanigans.

Specificity is important, its better to increase specificity before using exclusions, its not a rule of thumb but its something I should always try myself before exclusions and sometimes laziness sets in and out goes common sense. =)

wikipedia-dark.user.css
AfroThundr3007730: Hmm, every time I tried to put it back, `npm run clean` indented it again. I'll convert it to a regex then. How about this one?
```
regexp(https?:\/\/[^?]*((wikipedia|mediawiki|wikibooks|wikidata|wikinews|wikiquote|wikisource|wikiversity|wikivoyage|wiktionary)|(meta|commons|otrs-wiki|species|wikitech)\.wikimedia)\.org\/.+)
```

the-j0k3r:
LGTM didnt check myself the changes, but if you wanna merge'em go for it.

wikipedia-dark.user.css
the-j0k3r: >Hmm, every time I tried to put it back, npm run clean indented it again

That doesnt happen here at all. I always run it before doing a release.

> I'll convert it to a regex then. How about this one?

regex is Chinese to me, I admit defeat in there. So whatever works. =)

AfroThundr3007730: Regex it is then. :)

Issue #72, opened 2019-05-25
Mobile styling


Mottie:
This needs more work.

See: https://en.m.wikipedia.org

the-j0k3r:
Needs a maintainer for sure. /me taps out.

AfroThundr3007730:
Hmm styling the Minerva skin would be quite a bit of work, it seems. @the-j0k3r How much support does this style even have for non-standard (not Vector) skins anyway? Has it been tested with others, such as Monobook?

AfroThundr3007730:
Just answered my own question. Here's how the style fares with each of the default skins.

## Vector (desktop default):
![MediaWiki Vector](https://user-images.githubusercontent.com/7613032/58758096-e5e79180-8505-11e9-901c-e9f8a95f3cab.png)

## CologneBlue:
![MediaWiki CologneBlue](https://user-images.githubusercontent.com/7613032/58758101-fa2b8e80-8505-11e9-8666-321f81376076.png)

## Minerva (mobile default):
![MediaWiki Minerva](https://user-images.githubusercontent.com/7613032/58758109-10d1e580-8506-11e9-9556-00212d13c1b8.png)

## Modern:
![MediaWiki Modern](https://user-images.githubusercontent.com/7613032/58758113-1fb89800-8506-11e9-84f0-4657d3ef6a66.png)

## MonoBook:
![MediaWiki MonoBook](https://user-images.githubusercontent.com/7613032/58758121-2f37e100-8506-11e9-8c3a-710d1d67bd02.png)

## Timeless:
![MediaWiki Timeless](https://user-images.githubusercontent.com/7613032/58758136-4e367300-8506-11e9-994b-5c92ea943829.png)

AfroThundr3007730:
It would seem that aside from Vector (which we put the majority of our focus on), the Timeless skin fares the best as far as style coverage goes. I haven't found any obvious issues from my 10 second examination.

MonoBook and Modern aren't far behind, with some minor cleanup necessary to bring them up to standard, while CologneBlue will need a bit more work to do the same.

The Minerva skin is by far the worst of the bunch coverage-wise. I haven't dug too deeply into its structure yet, so I'm not sure how difficult it would be to style it without conflicting with elements of the Vector styling. It looks doable though..

the-j0k3r:
[...]

PR #74, opened 2019-06-02
Minor tweaks and template fixes

AfroThundr3007730:
This mainly fixes up some style issues in the VisualEditor, CiteIt, and other OOUI-based tools. Also made redlinks and diff numbers brighter and removed the background of h2 elements. The background for many of the talk page templates have been fixed to not be transparent as well.

Further commits exclude mobile pages from the site matching regex (until we can fix them properly), and fixes up some element styling on Wikidata. Also fixed WikiEd (yes, it still exists). More fixes for icon colors, and several icons were inlined as well.

Plus many smaller changes; the commit history will be the best source of change info.

Pretty much the last holdout is the syntax highlighting mode of the VisualEditor source editor, which for the life of me I can't seem to restore (it's all gray still). The class it toggles is `ve-ce-documentNode-codeEditor-webkit-hide`, but no amount of `!important` or `:not()` selectors seems to work.

This last part will be handled in a separate project.


AfroThundr3007730:
@the-j0k3r If you have any ideas re: syntax highlighting, that would be helpful. If not, I'll just let that bit stay as-is, though it's not ideal.

the-j0k3r:
> @the-j0k3r If you have any ideas re: syntax highlighting, that would be helpful. If not, I'll just let that bit stay as-is, though it's not ideal.

See relevant see https://github.com/StylishThemes/Wikipedia-Dark/pull/35#issuecomment-398160540

Though show me a link to page is better.

wikipedia-dark.user.css
```diff
@@ -122,7 +148,7 @@
   /*** Red links (non-existant pages) ***/
   div.vectorTabs li.new:not(.selected) a, .mw-plusminus-neg, a.new,
   #p-personal a.new {
-    color: #980000 !important;
+    color: #e33 !important;
```
the-j0k3r: Isnt this a bit too bright?

AfroThundr3007730: I was going for something that didn't just blend in with a dark theme, since redlinks are supposed to stand out a bit, and were harder to read with `#980000`. If there's another shade you'd like, I can swap it out.

the-j0k3r: I dont mind really, just that e33 may burn some retinas...

AfroThundr3007730: Yeah, we don't want it to be to obnoxious, how about a nice `#f11` then?

AfroThundr3007730: Done. I think `#f11` is a good compromise here.

AfroThundr3007730:
Yeah, it'll take a bit more work to fix up any collateral damage after removing that rule (which incidentally, is also the reason I'm having to use `!important` overrides all over the place).

I'm wondering if we might be better off removing the more generic ones and fixing up the regressions with a boatload of new selectors. This might need to happen anyway if we have to reorganize to support mobile styling (#72).

AfroThundr3007730:
Alright, the syntax highlighting (and the ensuing reorganization that will entail) will be tackled as a separate project. Another matter to consider is the site matching regex, which I'm going to update to exclude the mobile sites for now (pending a fix for #72):
```regex
https?:\/\/([a-zA-Z0-9\-]{2,}\.(wikipedia|mediawiki|wikibooks|wikidata|wikinews|wikiquote|wikisource|wikiversity|wikivoyage|wiktionary)|(meta|commons|otrs-wiki|species|wikitech)\.wikimedia)\.org\/.+
```
Sauce: https://regex101.com/r/zNHRno/2

Note that not all the domains on that linked list need be supported; only the ones that host a MediaWiki instance (which the regex currently covers, sans mobile). This should work fine, since the only single-character subdomain currently active is `m` for mobile. All other language code subdomains are at least two characters (indeed, there are no valid single-character language codes), so the regex should remain valid.

the-j0k3r:
>Alright, the syntax highlighting (and the ensuing reorganization that will entail) will be tackled as a separate project. 

Yes definitely. you have write permissions here, dont you? Small PRs like this could be merged also directly, unless its something youre not comfortable or sure about of course.

> Sauce: https://regex101.com/r/zNHRno/2

Mushroom and cream, coffee sauce. mmmm

AfroThundr3007730:
> Yes definitely. you have write permissions here, dont you? Small PRs like this could be merged also directly, unless its something youre not comfortable or sure about of course.

I have commit access, yes, but I generally will do long strings of changes in a branch, then do a PR so I can get a second set of eyeballs on them before [pushing to master](https://i.imgur.com/Tuh20lN.jpg).

the-j0k3r:
> ...so I can get a second set of eyeballs on them before [pushing to master](https://i.imgur.com/Tuh20lN.jpg).

or 3rd, cyclops style. You dont know, I may only have one eye. BAB knows.

AfroThundr3007730:
Well, that's about all I've got for this batch. There are still minor styling issues with the headers on the main page to resolve (after the restructuring some months back), but that can wait til another day.

@the-j0k3r If you find no problems with this PR, I'll go ahead and merge.

the-j0k3r:
I still think the red maybe too strong see https://en.wikipedia.org/w/index.php?title=Joko_Widodo&action=history and https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1

Also found other issues while testing style + this CSS, that this PR doesnt cover so consider this a **TODO** list.

- [ ] White textarea bg e.g https://www.mediawiki.org/w/index.php?title=Project:Requests/User_rights/Examknow&action=edit&section=1
- [ ] Black icons https://www.mediawiki.org/wiki/Special:Contributions and https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1
- [ ] Too light green/yellow bg and text + hovers in userSummaryButtons  https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1
- [ ] Unstyled side panel in https://www.mediawiki.org/wiki/Special:MyTalk https://www.mediawiki.org/wiki/Project_talk:Current_issues probably all same selectors.
- [ ] Button text too dim https://www.mediawiki.org/w/index.php?title=Special:Log/block&page=User%3

There are other likely. None of them have to be covered in this PR anyway.

AfroThundr3007730:
> I still think the red maybe too strong see https://en.wikipedia.org/w/index.php?title=Joko_Widodo&action=history and https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1

We've gone back and forth on this one. Do you have a decent shade of red that will not be to bright, yet still legible on a dark background? If we can find a good median, I'll put that in.

> * [x]  White textarea bg e.g https://www.mediawiki.org/w/index.php?title=Project:Requests/User_rights/Examknow&action=edit&section=1

I see a flash of unstyled content (FOUC) for a second, then the normal `#222` (or `#111` in syntax highlight mode) takes over. 

Hmm, pausing it in the debugger before the style catches up (and with JS disabled) I get a generic `textarea`, which only has `background-color: transparent !important`applied all the way up to `body` (which has `background-color: #222 !important`), so why is it coming out white?

No matter, added a rule for it.

> * [x]  Black icons https://www.mediawiki.org/wiki/Special:Contributions and https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1

It's always fun trying to change the icon color but keep the text color the same. I'm glad `filter()` exists. On that second link, the only unstyled elements were the down arrows, which are now fixed.

> * [x]  Too light green/yellow bg and text + hovers in userSummaryButtons  https://pl.wikipedia.org/w/index.php?title=Dyskusja:Grupy_zwi%C4%85zane_z_Wu-Tang&action=edit&redlink=1

Got em, although that one button in the top-right had its own inline style for some reason. There goes more `!important` to litter the stylesheet...

> * [x]  Unstyled side panel in https://www.mediawiki.org/wiki/Special:MyTalk https://www.mediawiki.org/wiki/Project_talk:Current_issues probably all same selectors.

Lol, I completely forgot about Flow, since we disabled it on enwiki like immediately after release. It's no longer forgotten and now conforms to the style.

> * [x]  Button text too dim https://www.mediawiki.org/w/index.php?title=Special:Log/block&page=User%3

Some fallout from the Wikidata label ixes, now re-fixed to ensure proper button contrast.

Find anything else needing fixes?

the-j0k3r:
Maybe this, though looking at that page the green needs a couple of ticks brighter if we use this red `#cc323d` ?

>Hmm, pausing it in the debugger before the style catches up (and with JS disabled) I get a generic textarea, which only has background-color: transparent !importantapplied all the way up to body (which has background-color: #222 !important), so why is it coming out white?

Its likely the useragent styles in FF its

```
textarea {
	-moz-appearance: textarea;
	margin-block-start: 1px;
	margin-block-end: 1px;
	border: 2px inset ThreeDLightShadow;
	padding-inline-start: 1px;
	padding-inline-end: 1px;
	background-color: -moz-Field;
	color: -moz-FieldText;
	font: medium -moz-fixed;
	text-rendering: optimizeLegibility;
	vertical-align: text-bottom;
	cursor: text;
	resize: both;
	-moz-user-select: text;
	white-space: pre-wrap;
	word-wrap: break-word;
	overflow-clip-box: content-box;
	text-align: initial;
	text-indent: initial;
	text-shadow: initial;
	text-transform: initial;
	word-spacing: initial;
	letter-spacing: initial;
}
```
Setting
```
textarea {
  -moz-appearance: none;
  background-color: transparent;
}
```

Makes it dark but bg should be at least `#111111` or `#181818`

AfroThundr3007730:
> Maybe this, though looking at that page the green needs a couple of ticks brighter if we use this red `#cc323d` ?

That's almost as intense as my original choice `#e33` (`#ee3333`). I'm thinking `#f11` (`#ff1111`) might be the best we can come up with (until we find someone great at picking these things).

> Makes it dark but bg should be at least `#111111` or `#181818`

I made it match the editor color `#222` (`#222222`) so there would be no difference while loading.

the-j0k3r:
> That's almost as intense as my original choice `#e33` (`#ee3333`). I'm thinking `#f11` (`#ff1111`) might be the best we can come up with (until we find someone great at picking these things).

#f11 is way brighter, maybe its your screen? It kills the balance between the green and red and its overwhelming in some pages.

AfroThundr3007730:
> Heres one `#b22` `#bb2222` and green `#6cc644` Those are from GitHub Dark

Hmm color pickers are handy. Now I have the proper vocabulary to articulate what I was getting at.

  - `#980000` (original color) was too dark and intense (saturation too high, value too low)
  - `#ee3333` (choice 1) ended up lighter, but with the saturation a bit low
  - `#ff1111` (choice 2) overcompensated on both saturation and value
  - `#cc323d` (suggestion 1) saturation is too low, I think, aiming towards pink
  - `#bb2222` (suggestion 2) pretty close actually, might go with that one
  - `#cc2222` (choice 3) slightly brighter, and seems like the best fit

I was aiming for a color that would be squarely in the "red" zone of red (if that makes any sense), so a fairly high saturation, while also being bright enough to read properly on a dark background, so it needs a higher value as well.

The last choice (`#c22`) results in a Saturation and Value > 80 but not too obnoxious, so I think it will be the best choice of the set.

AfroThundr3007730:
Btw, any strong opinion on the site matching regex -- namely, if it's optimized? 

  - Version 3: https://regex101.com/r/zNHRno/3
  - Version 4: https://regex101.com/r/zNHRno/4

It's a trade-off in efficiency vs legibility. Both are functionally identical.

the-j0k3r:
> Btw, any strong opinion on the site matching regex -- namely, if it's optimized?

None whatsoever, Regex is beyond me.

> I was aiming for a color that would be squarely in the "red" zone of red (if that makes any sense), so a fairly high saturation, while also being bright enough to read properly on a dark background, so it needs a higher value as well.

The point I was trying to get through and seemingly have not is, the color scheme has to fit together, so greens that show frequently against reds on same page dont look like came from a different style, with your choices I think leave the greens quite subdued when the red is way out of its range. If youre happy with that go with it. The alternative is to also brighten the green.
Personal preference is not towards piercing bright colors on a dark theme, my retinas get shocked =)

wikipedia-dark.user.css
```diff
@@ -283,6 +345,32 @@
     background-color: transparent !important;
     border-color: #555 !important;
   }
+  .mw-editfont-monospace {
+    background-color: #222;
+    color: #9a9a9a;
+  }
+  .userSummaryButtons a, #userSummaryButtons a {
+    background-color: #333;
```
the-j0k3r:
```suggestion
    background-color: #002800;
```

wikipedia-dark.user.css
```diff
@@ -283,6 +345,32 @@
...
+    border-color: #333;
```
the-j0k3r:
```suggestion
        border-color: #061;
```

wikipedia-dark.user.css
```diff
@@ -110,6 +110,52 @@
...
+  }
+  table.diff, td.diff-otitle, td.diff-ntitle {
+    background-color: #333;
+  }
+  .oo-ui-tool.oo-ui-widget-enabled > .oo-ui-tool-link:hover,
+  .oo-ui-tool.oo-ui-widget-enabled > .oo-ui-tool-link:active,
+  .oo-ui-tool.oo-ui-widget-enabled > .oo-ui-tool-link:active:focus,
+  .oo-ui-tool.oo-ui-widget-enabled.oo-ui-tool-active > .oo-ui-tool-link,
+  .oo-ui-tool.oo-ui-widget-enabled.oo-ui-popupToolGroup-active > .oo-ui-tool-link {
+    background-color: #333;
```
the-j0k3r: Merge table.diff, td.diff-otitle, td.diff-ntitle with the rule below.

wikipedia-dark.user.css
```diff
@@ -110,6 +110,52 @@
...
+  }
+  .oo-ui-checkboxInputWidget.oo-ui-widget-enabled [type="checkbox"]:checked + span,
+  .oo-ui-checkboxInputWidget.oo-ui-widget-enabled [type="checkbox"]:indeterminate + span {
+    background-color: #36c !important;
+    filter: none !important;
+  }
+  .oo-ui-menuOptionWidget.oo-ui-optionWidget-highlighted {
+    background-color: #333 !important;
```
the-j0k3r: this could be merged further up.

wikipedia-dark.user.css
```diff
@@ -110,6 +110,52 @@
...
+  }
+  .oo-ui-menuOptionWidget.oo-ui-widget-enabled.oo-ui-optionWidget {
+    color: #777;
+  }
+  .oo-ui-messageDialog-title, .oo-ui-messageDialog-message {
+    color: #9a9a9a;
+  }
+  #language-settings-dialog {
+    background-color: #222 !important;
```
the-j0k3r: all these three above same.

wikipedia-dark.user.css
```diff
@@ -283,6 +345,32 @@
...
+    color: #9a9a9a;
+  }
+  .userSummaryButtons a, #userSummaryButtons a {
+    background-color: #333;
+    border-color: #333;
+  }
+  .userSummaryButtons a:hover, #userSummaryButtons a:hover {
+    background-color: #444;
+    border-color: #444;
+  }
```
the-j0k3r:
```css
  .userSummaryButtons a, #userSummaryButtons a {
    background-color: #002800;
    border-color: #061;
    color: #aaa !important;
  }
  .userSummaryButtons a:hover, #userSummaryButtons a:hover {
    background-color: #002800;
    border-color: #061;
    filter: brightness(120%);
  }
```
Please use green colors to replace default green elements, same with other colors/elements and always when possible, sometimes its not possible but this isnt this case.

wikipedia-dark.user.css
```diff
@@ -283,6 +345,32 @@
...
+  #btn_opisy_zm_waxon_off {
+    background-color: #333 !important;
+    border-color: #333 !important;
+  }
```
the-j0k3r: Same with this button, the color scheme should be along 

```css
    background-color: #261d08 !important;
    border-color: #542 !important;
```

AfroThundr3007730:
> Edit (something along)
> 
> ```css
> .mw-plusminus-pos {
> 	color: #00ba00 !important;
> }
> ```

Hmm that looks alright, but the green seems a bit bright. It's like the green doesn't have to work as hard to match the red, so perhaps a bit more muted? I have another suggestion:

  - `#007f00` (current color)
  - `#00ba00` (your suggestion)
  - `#339900` (my suggestion)

I'm not invested in either solution so either one will work. :smiley:

the-j0k3r:
> I'm not invested in either solution so either one will work. smiley

You pick.

wikipedia-dark.user.css
```diff
@@ -520,13 +620,22 @@
...
-  .oo-ui-pendingElement-pending,
   #pt-notifications-alert .mw-echo-notifications-badge:before,
   img.mwe-math-fallback-image-display, #cnotice-toggle-icon {
     -webkit-filter: invert(100%) !important;
             filter: invert(100%) !important;
     background-color: transparent !important;
   }
+  .oo-ui-pendingElement-pending {
+    background-color: #ddd !important;
+    filter: invert(100%) !important;
```
the-j0k3r: Do we need to use this sort of solution?

wikipedia-dark.user.css
```diff
@@ -545,6 +654,12 @@
   #footer-icons img {
     filter: brightness(70%);
   }
+  #mw-indicator-mw-helplink a {
+    filter: invert(80%) hue-rotate(180deg);
```
the-j0k3r: Just apply straight colors to the selectors. Same with the rule below.

wikipedia-dark.user.css
```diff
@@ -559,6 +674,50 @@
...
+  .imeselector-toggle {
+    filter: brightness(200%);
+  }
+  .imeselector-menu {
+    filter: brightness(50%);
+    background-color: #222 !important;
+  }
+  .imeselector-menu .selectable-row:hover {
+    background-color: #333;
+  }
```
the-j0k3r: Again here above I see many selectors that need to be condensed into the same property rule.

wikipedia-dark.user.css
AfroThundr3007730: Implemented

wikipedia-dark.user.css
AfroThundr3007730: Implemented

wikipedia-dark.user.css
AfroThundr3007730: Two main thoughts here:

  - I'm trying to avoid combining rules with and without `!important`, since in some cases adding an unnecessary `!important` will cause unintentional style breakages elsewhere. There are several cases like that right now, and it's likely to grow in the future. If we can redo those general selectors so we don't need `!important` sprinkled all over the place to override them, more of these could be safely combined.
  - I'm also trying not to combine completely unrelated elements just because they happen to have the same rules at the moment (ex: a `button` border and the color of a `p` element). This can come back to bite us in the future when we need to change just one of them, and we're not sure if that selector was grouped with the others intentionally, or just because. I'll start combining the ones with similar roles/types/purposes, but there should be some separation. Grouping them together can allow for easier full-stylesheet changes if we want to change a particular color globally (or variables, but that's later).

wikipedia-dark.user.css
AfroThundr3007730: see above

wikipedia-dark.user.css
AfroThundr3007730: see above

wikipedia-dark.user.css
AfroThundr3007730: Well, to the best of my knowledge, the only way to make the animation on `oo-ui-pendingElement-pending` dark is to invert it. The problem is, it sometimes has child elements that need to be un-inverted (the echo notification icons, and the elements on the loading banner in the VisualEditor).

These child elements also clash with the generic icon inversion rules in some cases, and `filter: none !important` isn't helping either. If you have a better idea (besides reorganizing all those generic selectors causing trouble), I'm open to suggestions.

wikipedia-dark.user.css
AfroThundr3007730: Those are both images (set with `background-image`) so inverting them was the only way. In the particular case of the help link, there was no way to separate the text from the icon (usually, it's in the `::before` pseudoclass), meaning I had to fix the text color as well.

wikipedia-dark.user.css
AfroThundr3007730: Yes a lot of these will get combined.

wikipedia-dark.user.css
the-j0k3r: This is what I did to it https://gist.github.com/the-j0k3r/8f8823e6c5ac654875a47fe8107e12d8 then I link it in my own style, you feel free to import that image into a base64 or into the repo and use that instead.

wikipedia-dark.user.css
the-j0k3r: I need specific page links to look at what the problem is and work on solutions. But doing this is the last solution youde go for usually.

AfroThundr3007730:
Back on the topic of colors, I've converted them all to root variables. Here's what we ended up with:
```css
  :root {
    --base-color: /*[[base-color]]*/;
    --base-color-8: rgba(/*[[base-color-rgb]]*/, .8);
    --visited-color: /*[[visited-color]]*/;
    --bg-custom: /*[[bg-custom]]*/;
    --bg-selected: /*[[bg-selected]]*/;
    --bg-attachment: /*[[bg-attachment]]*/;
    --gray-1: #111; /* used 2x */
    --gray-2: #222; /* used 21x */
    --gray-3: #333; /* used 10x */
    --gray-4: #444; /* used 4x */
    --gray-5: #555; /* used 18x */
    --gray-7: #777; /* used 2x */
    --gray-a: #aaa; /* used 1x */
    --gray-c: #ccc; /* used 1x */
    --gray-d: #ddd; /* used 7x */
    --gray-e: #eee; /* used 1x */
    --white: #fff;  /* used 6x */
    --gray-18: #181818; /* used 1x */
    --gray-28: #282828; /* used 2x */
    --gray-2f: #2f4f4f; /* used 1x */
    --gray-69: #696969; /* used 1x */
    --gray-9a: #9a9a9a; /* used 5x */
    --gray-b9: #b9b9b9; /* used 1x */
    --gray-bc: #bcbcbc; /* used 1x */
    --blue-1: #36c;
    --blue-2: #182030;
    --blue-3: #212c43;
    --red-1: #c22;
    --red-2: #d33;
    --red-3: #722;
    --red-4: #ff4242;
    --green-1: #390;
    --green-2: #165316;
    --green-3: #061;
    --green-4: #002800;
    --brown-1: #321;
    --brown-2: #542;
    --brown-3: #261d08;
    --yellow-1: #b28200;
    --orange-1: #e53e00;
    --orange-2: #e08b26;
    --purple-1: #382b4a;
  }
```

I feel like some of those later grays (6-digit ones) could be consolidated or standardized to one of the more popular shades, to reduce the palette. I didn't count the usage of the other colors, although about half of them are only used in one rule (albeit often with a long list of selectors).

Any comments on the naming or implementation before I push these?

wikipedia-dark.user.css
AfroThundr3007730: Oh neat, I'll minimize and inline the SVG then (leaving a minor risk of eventual obsolescence).

the-j0k3r:
> Any comments on the naming or implementation before I push these?

Looks good I think.

wikipedia-dark.user.css
AfroThundr3007730: The `oo-ui-pendingElement-pending` class adds a "barbershop" stripe loading animation to an OOUI header.

You can see it in action when you click on the notifications box while logged in (as the notifications load):

  - [Capture 1](https://i.imgur.com/D9D28oG.mp4)

You can also see it briefly on the save dialog after making an edit, then clicking the preview button:

  - [Capture 2](https://i.imgur.com/VqxWP1M.mp4)

I added the class manually in these screenshots, but the real animation would happen for a second or two.

wikipedia-dark.user.css
AfroThundr3007730: Ok, most of the identical selectors have been combined, where it makes sense to do so.

wikipedia-dark.user.css
AfroThundr3007730: Here is the vendor style for that class:
```css
.oo-ui-pendingElement-pending {
  background-color: #eaecf0;
  background-image: -webkit-linear-gradient(135deg,#fff 25%,transparent 25%,transparent 50%,#fff 50%,#fff 75%,transparent 75%,transparent);
  background-image: -moz-linear-gradient(135deg,#fff 25%,transparent 25%,transparent 50%,#fff 50%,#fff 75%,transparent 75%,transparent);
  background-image: linear-gradient(135deg,#fff 25%,transparent 25%,transparent 50%,#fff 50%,#fff 75%,transparent 75%,transparent);
  background-size: 1.42857143em 1.42857143em;
  -webkit-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
     -moz-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
          animation: oo-ui-pendingElement-stripes 650ms linear infinite;
}
```
I guess we just need to invert the linear-gradient then. Lemme try something...

AfroThundr3007730: Oh snap! I got it. :smiley: Here's what I ended up with:
```css
  .oo-ui-pendingElement-pending {
    background-color: var(--gray-1);
    background-image: -webkit-linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
    background-image: -moz-linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
    background-image: linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
    background-size: 1.42857143em 1.42857143em;
    -webkit-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
       -moz-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
            animation: oo-ui-pendingElement-stripes 650ms linear infinite;
  }
```

AfroThundr3007730:
@the-j0k3r After making the linter happy, this looks good on my end. What do you think?

wikipedia-dark.user.css
the-j0k3r: In future please just drag screenshots into the reply area here instead of linking them from imgur, I rather not have to click and go elsewhere to see something related.

wikipedia-dark.user.css
```diff
@@ -520,20 +631,26 @@
...
+  .oo-ui-pendingElement-pending {
+    background-color: var(--gray-1);
+    background-image: -webkit-linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
+    background-image: -moz-linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
+    background-image: linear-gradient(135deg, black 25%, transparent 25%, transparent 50%, black 50%, black 75%, transparent 75%, transparent);
+    background-size: 1.429em 1.429em;
+    -webkit-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
+       -moz-animation: oo-ui-pendingElement-stripes 650ms linear infinite;
+            animation: oo-ui-pendingElement-stripes 650ms linear infinite;
+  }
```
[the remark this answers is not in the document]
the-j0k3r: any properties like size and animation etc that exist in vendor shouldn't need to also be in style.

wikipedia-dark.user.css
AfroThundr3007730: A screenshot isn't the best medium for capturing an animation though. I couldn't figure out how to make them show as a GIF instead of a video, so it wouldn't embed. I could just upload the video files here directly, I guess.

wikipedia-dark.user.css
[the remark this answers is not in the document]
AfroThundr3007730: You're right, the last four properties don't need to be in there. I'll drop em.

AfroThundr3007730:
Think I've covered everything needed for this batch (and then some). Several days of testing have found no other issues, so I'll go ahead and merge this.

PR #78, opened 2019-08-15
Begin dark theme skeleton for mobile style

### LINKED ISSUE #72: Mobile styling

AfroThundr3007730:
<s>First whack at</s> Fixes #72

I'm attempting to reuse the existing color palette as much as possible, though at the moment some of the elements won't exactly match their desktop counterparts. I can hunt those down and fix them later.

<s>And yes, @the-j0k3r, I'll be merging all the selectors with duplicate declarations later (I can already hear you saying that). I'm still playing unstyled element whack-a-mole for now.</s> :smiley:

After thorough testing, the new mobile style (for Minerva) should now be complete:tm:.

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -910,6 +910,7 @@
     --orange-2: #e08b26;
     --purple-1: #382b4a;
     --purple-2: #768;
+    --cyan-1: #194d41;
     --ores-1: #4d3a00;
     --ores-2: #4d2900;
     --ores-3: #4d2004;
@@ -1389,8 +1390,27 @@
     background-color: var(--gray-2);
     border-color: var(--gray-7);
   }
+  .mw-rollback-link, .mw-tag-marker {
+    background-color: var(--gray-2);
+  }
   .mw-rcfilters-ui-highlights-color-none {
     background-color: var(--gray-7);
+  }
+  .mw-rcfilters-ui-highlights-color-c1,
+  .mw-rcfilters-ui-tagItemWidget-highlight[data-color="c1"]::before,
+  .mw-rcfilters-ui-changesListWrapperWidget li.mw-rcfilters-highlight-color-c1,
+  .mw-rcfilters-ui-changesListWrapperWidget .mw-rcfilters-highlight-color-c1 tr:first-child,
+  .mw-rcfilters-ui-changesListWrapperWidget tr.mw-rcfilters-highlight-color-c1.mw-rcfilters-ui-highlights-enhanced-toplevel:not(.mw-rcfilters-ui-changesListWrapperWidget-enhanced-grey) td:not(:nth-child(-n+2)),
+  .mw-rcfilters-ui-changesListWrapperWidget tr.mw-rcfilters-highlight-color-c1.mw-rcfilters-ui-highlights-enhanced-nested:not(.mw-rcfilters-ui-changesListWrapperWidget-enhanced-grey) td:not(:nth-child(-n+4)) {
+    background-color: var(--blue-2);
+  }
+  .mw-rcfilters-ui-highlights-color-c2,
+  .mw-rcfilters-ui-tagItemWidget-highlight[data-color="c2"]::before,
+  .mw-rcfilters-ui-changesListWrapperWidget li.mw-rcfilters-highlight-color-c2,
+  .mw-rcfilters-ui-changesListWrapperWidget .mw-rcfilters-highlight-color-c2 tr:first-child,
+  .mw-rcfilters-ui-changesListWrapperWidget tr.mw-rcfilters-highlight-color-c2.mw-rcfilters-ui-highlights-enhanced-toplevel:not(.mw-rcfilters-ui-changesListWrapperWidget-enhanced-grey) td:not(:nth-child(-n+2)),
+  .mw-rcfilters-ui-changesListWrapperWidget tr.mw-rcfilters-highlight-color-c2.mw-rcfilters-ui-highlights-enhanced-nested:not(.mw-rcfilters-ui-changesListWrapperWidget-enhanced-grey) td:not(:nth-child(-n+4)) {
+    background-color: var(--cyan-1);
   }
   .mw-rcfilters-ui-highlights-color-c3,
   .mw-rcfilters-ui-tagItemWidget-highlight[data-color="c3"]::before,
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -1389,6 +1389,10 @@
     background-color: var(--gray-2);
     border-color: var(--gray-7);
   }
+  .mw-rcfilters-ui-highlights-color-none {
+    background-color: var(--gray-7);
+  }
+  .mw-rcfilters-ui-highlights-color-c3,
   .mw-rcfilters-ui-tagItemWidget-highlight[data-color="c3"]::before,
   .mw-rcfilters-ui-changesListWrapperWidget li.mw-rcfilters-highlight-color-c3,
   .mw-rcfilters-ui-changesListWrapperWidget .mw-rcfilters-highlight-color-c3 tr:first-child,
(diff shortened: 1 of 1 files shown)


AfroThundr3007730:
@the-j0k3r Quick question: When browsing [Special:RecentChanges](https://en.m.wikipedia.org/wiki/Special:RecentChanges?hidebots=1&limit=100&days=7&enhanced=1&damaging__maybebad_color=c3&damaging__likelybad_color=c4&damaging__verylikelybad_color=c5&urlversion=2), you may notice that there's no left and right padding on the `li` elements, which causes them to run to the edge of the display (on a mobile screen), and in some cases causes the contents to overflow a bit (especially noticeable with the ORES highlighting enabled).

I was going to fix this by adding some padding, and fixing the bullet point location:

```css
.mw-rcfilters-ui-changesListWrapperWidget-highlighted ul li {
  padding: .5em;
}
.mw-rcfilters-ui-changesListWrapperWidget .mw-rcfilters-ui-highlights > div {
  margin-right: .5em;
}
```

Normally, we don't want to move elements around, and just set colors only, but this seems like a straightforward fix. Thoughts?

the-j0k3r:
> @the-j0k3r Quick question: When browsing [Special:RecentChanges](https://en.m.wikipedia.org/wiki/Special:RecentChanges?hidebots=1&limit=100&days=7&enhanced=1&damaging__maybebad_color=c3&damaging__likelybad_color=c4&damaging__verylikelybad_color=c5&urlversion=2), you may notice that there's no left and right padding on the `li` elements, which causes them to overflow the list boundary a bit (especially noticeable with the ORES highlighting enabled).

Can you show me a screenshot marking the issue, navigating there its nothing stands out.

AfroThundr3007730:
A screenshot of the behavior (mobile emulated):

Screen Shot 2019-08-16 at 07 10 13

Aside from the bullets, the `li` elements have no left/right padding (and perhaps too much top).

I think in this case adding the padding rules sould be fine, since the mobile RecentChanges hasn't gotten much attention by the devs for now.

the-j0k3r:
Just add a note above the CSS that fixes this issue upstream. and make that commit separate from the others perhaps.

wikipedia-dark.user.css
```diff
@@ -1319,6 +1338,19 @@
   .mw-rcfilters-ui-changesListWrapperWidget tr.mw-rcfilters-highlight-color-c3.mw-rcfilters-highlight-color-c4.mw-rcfilters-highlight-color-c5.mw-rcfilters-ui-highlights-enhanced-nested:not(.mw-rcfilters-ui-changesListWrapperWidget-enhanced-grey) td:not(:nth-child(-n+4)) {
     background-color: var(--ores-3);
   }
+  .mw-rcfilters-ui-changesListWrapperWidget {
```
the-j0k3r: note? separate commit?

AfroThundr3007730: I'm fixing that

AfroThundr3007730:
@the-joker Just rebased onto the latest release. I'll combine the selectors in a bit. In the meantime, if you're not busy, feel free to give it a spin and see if you can find any issues with the mobile theme.

the-j0k3r:
>  if you're not busy, feel free to give it a spin and see if you can find any issues with the mobile theme.

I found some.

* border colors too bright in places, especially tags
* links not base-color in few places
* hover states are broken i.e it has background on hover and default doest.(view new changes since....)
* brown colors background should be `#321` - `#332211` or anything around that darker `#33220a`
* unfolded burger menu looks a little harsh.

And thats just one page

Screenshot_2019-08-18 Recent changes - Wikipedia

I wont have much time in the next few days so I wont have time for spinning around =)

What I usually do, is switch the style off and on and try to match the default look as closely as possible, only dark.

Good effort though, Ill only see this in testing as I dont use mobile at all.

AfroThundr3007730:
Thanks for the feedback!

> * border colors too bright in places, especially tags

Yeah I missed those tag borders. I'm aiming to make pretty much all of them `--gray-7`. The `--gray-5` used in the desktop theme seems a bit dark, and this theme is going to be on smaller screens, so I think the brighter color works better here.

That's the same reason most of the foreground colors are `--gray-c` instead of `--gray-9a`, with secondary elements getting the 9a treatment (I'm still catching and fixing those).
 
> * links not base-color in few places

Other than redlinks, and a few other classes seen in article prose, they should all be `--base-color`. I'll keep hunting down the outliers.

> * hover states are broken i.e it has background on hover and default doest.(view new changes since....)

Yeah, I'll keep knocking those out as I find them. I'm doing a lot of QC from my actual phone, so I often miss hover states. Testing again from my desktop is revealing more of them.

> * brown colors background should be `#321` - `#332211` or anything around that darker `#33220a`

Yeah, I'm not the best at picking new colors (remember the redlinks discussion?), and being a bit colorblind (deuteranopia) doesn't help. Working with a preexisting palette is easier.

I got the current ORES colors by darkening the vendor colors until the Value (HSV) was below 20%. I just need three progressively darker browns (and a blue, and a cyan) that don't clash with the rest of the page.

> the colors used for bg especially the darker seems to fall outside our palette.

Which ones? Aside from the ORES highlights, I haven't introduced any major new colors.

> * unfolded burger menu looks a little harsh.

You mean the main menu in the top left? Could you define "harsh" exactly?

Speaking of menus, the filter menus in RecentChanges are lighter than their desktop counterparts (which were reduced to a flat `--gray-2`) and seem to look better this way. Those and the notifications menu are sporting a `--gray-4` with a `--gray-3` highlight, but that can be reversed if you think they should be a bit darker.

> I wont have much time in the next few days so I wont have time for spinning around =)

No problem, there's no rush on pushing this, and I want to iron out more bugs before pushing ahead. Anyone who really wants this now can just directly install this branch anyway.

> What I usually do, is switch the style off and on and try to match the default look as closely as possible, only dark.

That's my general aim as well. This mostly involves trying to keeping minor elements from disappearing (many of our backgrounds match vendor foregrounds), and trying to integrate everything properly with the dark theme.

One last holdout is wikitables with highlighted cells. They're one of the few times I've used `!important` in the mobile style. I'm debating whether to make specific rules for each color like the desktop style did, or just a catch-all for any cell or row with a background property defined (which would still cause some interesting results in corner cases, but you can't win em all).

> Good effort though, Ill only see this in testing as I dont use mobile at all.

No problem, a second set of eyes is always helpful, even if temporary. :smiley:

the-j0k3r:
> Yeah, I'll keep knocking those out as I find them. I'm doing a lot of QC from my actual phone, so I often miss hover states. Testing again from my desktop is revealing more of them.

I also use a user-agent switcher, I find it works better in some cases than browser built in. Using an actul.

> You mean the main menu in the top left? Could you define "harsh" exactly?

Theres too much contrast between too bright borders, background and menu sections, the default this is much more tame, we shpould always emulate the default as closely as possible.


> Yeah I missed those tag borders. I'm aiming to make pretty much all of them `--gray-7`. The `--gray-5` used in the desktop theme seems a bit dark, and this theme is going to be on smaller screens, so I think the brighter color works better here.
> 
> That's the same reason most of the foreground colors are `--gray-c` instead of `--gray-9a`, with secondary elements getting the 9a treatment (I'm still catching and fixing those).

This is dealers choice, as I dont use mobile, It doesnt affect me really,

> No problem, a second set of eyes is always helpful, even if temporary. smiley

We use different stylistic approaches to styling some elements, as I dont visit these pages with an actual mobile, my actual perception may be skewed of what works in a small screen, my rule of thumb leans to consistency in general.

Good hunting. Happy to lend my eyes in any case.

AfroThundr3007730:
> * brown bg are still to light

Think I found a better set of replacement browns.

> * toolitips text color and arrow (when hovering filter tags)

Got em, and all of the variations (I think).

Also decided to reverse the menu colors to `--gray-3` with `--gray-4` hover/highlight.

AfroThundr3007730:
Ok, table colors were the last major holdout I could find, and that's just been squashed.

After rebasing this onto `master` I think this will be ready to merge.

PR #110, opened 2020-01-27
Fix for missing background color

theslimshaney:
Special message on the home page has a different color and isn't being changed to dark theme.

Default
image

Current
image

Proposed
image

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -2604,7 +2604,7 @@ regexp("https?:\/\/wiki\.(rpcs3\.net|(archlinux|mozilla)\.(org|jp))\/.*$") {
   .mw-rcfilters-ui-filterMenuOptionWidget.oo-ui-flaggedElement-muted:not(.oo-ui-optionWidget-selected) {
     background-color: var(--gray-2);
   }
-  div[style*="background-color:#f9f9f9"] {
+  div[style*="background-color:#f9f9f9"], div[style*="background-color:#fffaf5"] {
     background-color: var(--gray-2) !important;
   }
   table.fmbox-warning {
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -2604,7 +2604,7 @@ regexp("https?:\/\/wiki\.(rpcs3\.net|(archlinux|mozilla)\.(org|jp))\/.*$") {
   .mw-rcfilters-ui-filterMenuOptionWidget.oo-ui-flaggedElement-muted:not(.oo-ui-optionWidget-selected) {
     background-color: var(--gray-2);
   }
-  div[style*="background-color:#f9f9f9"] {
+  div[style*="background-color:#f9f9f9"], div[style*="background-color:#fffaf5"] {
```
the-j0k3r: It would be nice if that border was also fixed =)

wikipedia-dark.user.css
```diff
@@ -2604,8 +2604,9 @@ regexp("https?:\/\/wiki\.(rpcs3\.net|(archlinux|mozilla)\.(org|jp))\/.*$") {
...
     background-color: var(--gray-2) !important;
+    border-color: rgb(85, 85, 85) !important;
```
theslimshaney: I think this could fix it, but not sure as the message isn't showing up for me any more.

wikipedia-dark.user.css
theslimshaney: @the-j0k3r added something that I think could work, but can't test it

wikipedia-dark.user.css
the-j0k3r: 2 things.

1) Dont add borders to rules that dont have a border. In this case looks for a existing rule that matches background `var(--gray-2)` and has a border-color of `var(--gray-4)` or `var(--gray-5)`
2) Use the existing variables for colors. In this case it would be`var--gray-4) or var(--gray-5)`

Also, cant really see this either, Ive cleared my cookies and still cant see this message.

the-j0k3r:
There has been a technical update that removes the inline styles in the homepage, so this is now likely to be no longer needed.

I mean its probably needed but with the correct selectors instead of overriding the inline styles.

Ill close this for now, for more info see #112

AfroThundr3007730:
FWIW, that banner was produced by [Template:Main Page banner](https://en.wikipedia.org/wiki/Template:Main_Page_banner), which used one-off content and styles anyway, so there was no guarantee that the next person who configured it would use a similar styling.

PR #143, opened 2020-09-13
Fixes with weather tables

Atexor:
Fixed light gray text on light background in various weather tables. 

Example 1:
Before: 
After: 

Example 2:
Before: 
After: 

Fixed issue with dark color and dark background on Russian version of Wikipedia
Before: 
After: 

I hope this it will be enough. The color of these tables should be more unified...


AfroThundr3007730:
An improvement overall, at the cost of reduced contrast for the dark color backgrounds. I don't see a way around that though without writing a whole series of rules for every background or invoking javascript to fix it.

the-j0k3r:
Thanks for the PR

PR #154, opened 2021-01-15
Fix 20-year birthday banner

TotalCaesar659:
P.S. Happy anniversary, Wikipedia!

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -2563,6 +2563,10 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   div[style*="background-color: #FFE" i] div h2 span#Add_your_project {
     background-color: var(--brown-3);
   }
+  .mw-parser-output #mp-banner {
+    background-color: var(--brown-1);
+    border-color: var(--brown-2);
+  }
   .mw-parser-output .nmbox-header, .mw-parser-output .nmbox {
     background-color: var(--brown-3);
     border-color: var(--brown-2);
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -3470,10 +3470,14 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
...
   /*** main page ***/
-  .mw-parser-output #mp-topbanner, .mw-parser-output #mp-banner {
+  .mw-parser-output #mp-topbanner {
     border-color: var(--gray-5);
     background-color: var(--gray-2);
   }
+  .mw-parser-output #mp-banner {
+    border-color: var(--yellow-2);
+    background-color: var(--yellow-1);
+  }
```
the-j0k3r, line 3480: great were nearly there =) by yellow I meant brwon which is the standard goto color for yellow replacements.
```suggestion
  .mw-parser-output #mp-banner {
    background-color: var(--brown-1);
    border-color: var(--brown-2);
  }
```

Also this needs to be moved around to this area https://github.com/StylishThemes/Wikipedia-Dark/blame/master/wikipedia-dark.user.css#L2549

the-j0k3r, line 3480: great were nearly there =) by yellow I meant brwon which is the standard goto color for yellow replacements.
```suggestion
  .mw-parser-output #mp-banner {
    background-color: var(--brown-1);
    border-color: var(--brown-2);
  }
```

Also this needs to be moved around to this area https://github.com/StylishThemes/Wikipedia-Dark/blame/master/wikipedia-dark.user.css#L2549

PR #160, opened 2021-01-24
Fix keyboard keys again

TotalCaesar659:
https://ru.m.wikipedia.org/wiki/Print_Screen

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -4558,10 +4558,11 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
   .client-js body.mw-rcfilters-ui-loading .mw-changeslist {
     opacity: .7;
   }
-  kbd.keyboard-key {
+  kbd.keyboard-key, kbd {
     background: var(--gray-18) !important;
     border-color: var(--gray-5) !important;
     color: var(--gray-9a) !important;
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -4558,10 +4558,11 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
   .client-js body.mw-rcfilters-ui-loading .mw-changeslist {
     opacity: .7;
   }
-  kbd.keyboard-key {
+  kbd.keyboard-key, kbd {
     background: var(--gray-18) !important;
     border-color: var(--gray-5) !important;
     color: var(--gray-9a) !important;
+    -webkit-box-shadow: .1em .2em .2em var(--gray-4) !important;
```
Mottie, line 4565: Please change this to `box-shadow` so it works in all browsers

Mottie (changes requested):

wikipedia-dark.user.css
the-j0k3r: also make ru changes into ru section. other section may not reach this and vice versa

TotalCaesar659: Where I need to move it? Tried section ru.wikipedia.org (line 6307) but it doesn't apply.

the-j0k3r: its because youre modding ru mobile, so create one?

TotalCaesar659: Got it.

Issue #156, opened 2021-01-18
Hover tooltips have white background


TotalCaesar659:
* **Browser**: Chromium 87.0.4280.67
* **Operating System**: Lubuntu 20.04.1
* **Link to page with the issue**: https://en.wikipedia.org/wiki/Main_Page
* **Screenshot**:

![1](https://user-images.githubusercontent.com/14265316/104937735-2fc50a00-59bf-11eb-95cf-b06629fb2fb3.png)

MLavrentyev:
Same on Firefox 84.0.2 on both macOS Big Sur and Catalina

TheNormalestPeep:
Also happens on the latest version of Chrome for Windows 10

the-j0k3r:
Its not OS or browser issue related, they updated their css and the classes no longer match.

jason-wihardja:
I've made a pull request to fix this issue. This is how the page now looks in my browser.

![image](https://user-images.githubusercontent.com/3488460/105580923-eb23df00-5dc1-11eb-9a1d-0e0c63b1bfc7.png)

Kindly review the code. Thanks!

the-j0k3r:
fixed then by https://github.com/StylishThemes/Wikipedia-Dark/commit/8cd3388d0c51419c91f5c3fee923ceed2f1a041e

TotalCaesar659:
I hate to re-open issues, but it's not fixed on mobile pages:
https://en.m.wikipedia.org/wiki/Main_Page
![1](https://user-images.githubusercontent.com/14265316/105633541-35f93180-5e6a-11eb-8000-b8c88d9d565e.png)

jason-wihardja:
@TotalCaesar659 can you check using the code in the pull request and see whether this one fixes it for you? Thanks

TotalCaesar659:
@jason-wihardja LGTM.
![1](https://user-images.githubusercontent.com/14265316/105726644-56dd8780-5f3b-11eb-831d-88b1d4433ba0.png)

PR #161, opened 2021-01-24
Mobile Page Fixes for Popups

jason-wihardja:
This should do the trick for #156

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -1276,8 +1276,10 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   .messagebox, .errorbox, .warningbox, .successbox,
   .suggestions-special .special-label, .reference-text,
   .mwe-popups .mwe-popups-container, #viewsourcetext {
-    background-color: var(--gray-2) !important;
     color: var(--gray-c);
+  }
+  .mwe-popups .mwe-popups-container {
+    background-color: var(--gray-2) !important;
   }
   .mwe-popups .mwe-popups-container .mwe-popups-discreet {
     background-color: var(--white) !important;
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -1276,9 +1276,11 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   .messagebox, .errorbox, .warningbox, .successbox,
   .suggestions-special .special-label, .reference-text,
   .mwe-popups .mwe-popups-container, #viewsourcetext {
-    background-color: var(--gray-2) !important;
```
Mottie, line 1279: Why is this being removed for `.messagebox, .errorbox, ...`? It would be better to remove the `.mwe-popups .mwe-popups-container` selector from this group than to break the group.

wikipedia-dark.user.css
```diff
@@ -4192,6 +4194,26 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
   tr[bgcolor], td[bgcolor] {
     background-color: var(--gray-28);
   }
+  /* popups fixes */
+  .mwe-popups .mwe-popups-container {
+    background-color: var(--gray-2) !important;
+  }
+  .mwe-popups .mwe-popups-container .mwe-popups-discreet {
```
Mottie, line 4201: Please target the image `.mwe-popups-thumbnail` instead of the wrapper, when you target the wrapper, it sometimes leaves a large area of white background exposed.

Mottie (changes requested):
Thanks for helping out!

wikipedia-dark.user.css
jason-wihardja: The code was like that before my previous changes, so this was like a revert to the previous code. Apparently, we still need the color style for this tag

wikipedia-dark.user.css
jason-wihardja: Let me check that one. The purpose of this code is to apply white background on images so svgs would still be readable

wikipedia-dark.user.css
jason-wihardja: You may take a look at my previous commit here: https://github.com/StylishThemes/Wikipedia-Dark/commit/8cd3388d0c51419c91f5c3fee923ceed2f1a041e. So in this commit, I reverted the change and apply it on a separate selector below because I wasn't sure how it would behave on the other selectors in the group.

the-j0k3r:
If you mod the commit title to include fixes #156 the relevant issue will close automatically once this is merged.

jason-wihardja:
Ok, I've renamed the commit title to include such message

the-j0k3r:
Another suggestion is also to create a general mobile @-moz--doc section so that mobile and desktop fixes dont decide to conflict.

Issue #151, opened 2020-11-06
Maintainers Needed


the-j0k3r:
TLDR

I no longer have time to maintain 3rd party projects like this one.

If you have or find issues with this style (or any other from StylishThemes) consider making a contribution with fixes. Else, you may find your issues no longer will be fixed like they have been until this day.

Been a wild run. But all good things come to an end.

AfroThundr3007730:
@the-j0k3r I can take over this repo to keep the major issues fixed up. I've been busy lately, but I'll still be around.

tambry:
@the-j0k3r Thanks for you service!

the-j0k3r:
Thanks, hopefully in time things will return to normal.

Hasimir:
Entirely understandable, I've also been _far_ too caught up in other things to be able to step up here.

I removed `npm` years ago for security reasons (keeping it around while working on crypto is foolish), so running the full test suite wouldn't really be feasible anymore anyway.

I admire your optimism on things returning to "normal" given what we've all seen this year, though. ;)

o-r-i-z-u-r-u:
Thank for you everything you have contributed

PR #162, opened 2021-01-27
Miscellaneous white background fixes

jason-wihardja:
This is to fix several more issues with white background. All examples are taken from the page https://en.wikipedia.org/wiki/Riemann_hypothesis

- Commit https://github.com/StylishThemes/Wikipedia-Dark/commit/8304d001544a6600daa7e589ac7f038b5842b12e fixes white background for tables on desktop

    image

- Commit https://github.com/StylishThemes/Wikipedia-Dark/commit/335aebb4d50056455c16b6be763ef65e185140c8 fixes white background on mobile quotebox

    White Quotebox

- Commit https://github.com/StylishThemes/Wikipedia-Dark/commit/958f8117fdafded492ed7540b6e5a71419f16006 fixes white background on mobile lazy loading placeholder image

    White Lazy

- Commit https://github.com/StylishThemes/Wikipedia-Dark/commit/e04f567f362d6a8e7f551e516b47acee2d4b9587 is to remove redundant style (has been mentioned on line 2870 of the code). In addition, background color for mobile is also adjusted so that texts on svg image displays correctly

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -1276,8 +1276,10 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   .messagebox, .errorbox, .warningbox, .successbox,
   .suggestions-special .special-label, .reference-text,
   .mwe-popups .mwe-popups-container, #viewsourcetext {
-    background-color: var(--gray-2) !important;
     color: var(--gray-c);
+  }
+  .mwe-popups .mwe-popups-container {
+    background-color: var(--gray-2) !important;
   }
   .mwe-popups .mwe-popups-container .mwe-popups-discreet {
     background-color: var(--white) !important;
@@ -4191,6 +4193,26 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
   }
   tr[bgcolor], td[bgcolor] {
     background-color: var(--gray-28);
+  }
+  /* popups fixes */
+  .mwe-popups .mwe-popups-container {
+    background-color: var(--gray-2) !important;
+  }
+  .mwe-popups .mwe-popups-container .mwe-popups-discreet {
+    background-color: var(--white) !important;
+  }
+  .mwe-popups .mwe-popups-container a {
+    color: var(--gray-c) !important;
+  }
+  .mwe-popups-extract::after {
+    background-image: linear-gradient(to right, transparent, var(--gray-2) 50%) !important;
+  }
+  .mw-mmv-ttf-ellipsis::before {
+    background-image: linear-gradient(to right, transparent, var(--gray-2)) !important;
+  }
+  .mwe-popups-extract .mwe-popups-fade,
+  .mw-mmv-permission-box .mw-mmv-permission-text .mw-mmv-permission-text-fader {
+    background-image: linear-gradient(to bottom, transparent, var(--gray-2)) !important;
   }
   /* these general selectorss have to go when adding not remove all */
   td[style*="background"] {
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -2853,6 +2853,7 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   .mw-echo-ui-notificationBadgeButtonPopupWidget-popup > .oo-ui-popupWidget-popup > .oo-ui-popupWidget-head,
   .mw-echo-ui-notificationItemWidget:last-child {
     border-color: var(--gray-5);
+    background-color: var(--gray-2);
   }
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -1974,10 +1974,6 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   td[style*="background: #F5F5FF"] {
     background-color: var(--gray-28) !important;
   }
-  html .thumbimage {
-    background-color: var(--gray-28);
-    border-color: var(--gray-5);
-  }
```
the-j0k3r, line 1978: you sure that archlinux/mozilla doesnt have this? It woulnt be here otherwise.

wikipedia-dark.user.css
```diff
@@ -2853,6 +2853,7 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
   .mw-echo-ui-notificationBadgeButtonPopupWidget-popup > .oo-ui-popupWidget-popup > .oo-ui-popupWidget-head,
   .mw-echo-ui-notificationItemWidget:last-child {
     border-color: var(--gray-5);
+    background-color: var(--gray-2);
```
the-j0k3r, line 2852: I would instead be more specific and add bg color to the specific selector needed instead of all rule.

It will minimize the chances of breaking N over fixing x

wikipedia-dark.user.css
```diff
@@ -4192,6 +4194,26 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
...
+  .mwe-popups-extract::after {
+    background-image: linear-gradient(to right, transparent, var(--gray-2) 50%) !important;
+  }
+  .mw-mmv-ttf-ellipsis::before {
+    background-image: linear-gradient(to right, transparent, var(--gray-2)) !important;
+  }
+  .mwe-popups-extract .mwe-popups-fade,
+  .mw-mmv-permission-box .mw-mmv-permission-text .mw-mmv-permission-text-fader {
+    background-image: linear-gradient(to bottom, transparent, var(--gray-2)) !important;
+  }
```
the-j0k3r, line 4222: I would suggest minimize the use of !important unless its absolutely necessary for the override.

Also both popups are missing a border/box shadow where the arrow side of the popup.

the-j0k3r (changes requested):

wikipedia-dark.user.css
jason-wihardja: Oops, seems like I made a mistake here, I was meant for the selector below this. As you can see in line 2867 here: https://github.com/StylishThemes/Wikipedia-Dark/pull/162/files/e04f567f362d6a8e7f551e516b47acee2d4b9587#diff-b52ab5bfdb3a915b5378cc6e1f512df5667751a1d00a9e55ef613eac36ef46a7R2866, the selector already contains `html .thumbimage`. So my intention was to move the deleted one from https://github.com/StylishThemes/Wikipedia-Dark/pull/162/files/e04f567f362d6a8e7f551e516b47acee2d4b9587#r565533042 into here. Let me fix it

wikipedia-dark.user.css
jason-wihardja: This is redundant. You can check line 2868. It is already there. Also, the background-color actually makes the text on svg become unreadable. So I removed it.

wikipedia-dark.user.css
jason-wihardja: Ok, let me fix it

wikipedia-dark.user.css
jason-wihardja: For this part of the code, I actually just copy and pasted it from the desktop version. I notice that the desktop version also have the same issue. So maybe for this PR, I include only these changes to fix white background, and perhaps when I had time I would look into the missing border and create separate PR for it

wikipedia-dark.user.css
the-j0k3r: I dont mind if this breaks something, Im sure its bound to break something you aren't intending with the removal, also I dont really have time to go investigate which isnt helpful but there is #151 where its clear I dont have time for these side things =(

so @AfroThundr3007730 if he has time please, if not idk.

wikipedia-dark.user.css
the-j0k3r: Sure but it is already an issue on desktop version, just because something has existed for a long time, it shouldnt be perpetuated if not absolutely necessary.

AfroThundr3007730: So after testing, just the above vendor-provided selector is necessary for that last rule, and `!important` can be removed from all of them. Everything else is good.

AfroThundr3007730 (changes requested):

AfroThundr3007730 (approved):

PR #164, opened 2021-03-09
Fix contrast issue with conjugation tables in wiktionary

oowekyala:
I found a problem with some tables on the wiktionary: the text is too light and unreadable. For instance on [this page](https://en.wiktionary.org/wiki/limpiar):

Capture d’écran de 2021-03-09 09-08-34

With this PR it looks like so: 
Capture d’écran de 2021-03-09 09-08-54

Maybe a better solution would be to darken the background, but I don't know how to do that (the background is defined in a `style` attribute)

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -98,6 +98,7 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
     --brown-4: #4b380d;
     --yellow-1: #b28200;
     --yellow-2: #ffdd7f;
+    --yellow-3: #a88b2c;
     --orange-1: #e53e00;
     --orange-2: #e08b26;
     --purple-d: #241738;
@@ -6546,8 +6547,48 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
   .list-switcher-element a[role="button"] {
     color: var(--base-color) !important;
   }
-  .NavContent > table > tbody > tr > th {
-    color: var(--gray-5);
+  tr > *[style*="background:#c0cfe4"],
+  tr > *[style*="background:#9999DF"] {
+    background-color: var(--blue-4) !important;
+  }
+  tr > *[style*="background:#a0ade3"],
+  tr > *[style*="background:#ccccff"] {
+    background-color: var(--blue-3) !important;
+  }
+  tr > *[style*="background:#e2e4c0"],
+  tr > *[style*="background:#d0cfa4"] {
+    background-color: var(--yellow-1) !important;
+  }
+  tr > *[style*="background:#c0d8e4"] {
+    background-color: var(--yellow-3) !important;
+  }
+  tr > *[style*="background:#e0e0b0"] {
+    background-color: var(--pink-1) !important;
+  }
+  tr > *[style*="background:#c0e4c0"],
+  tr > *[style*="background:#99cc99"],
+  tr > *[style*="background:#99CC99"] {
+    background-color: var(--green-3) !important;
+  }
+  tr > *[style*="background:#cfedcc"] {
+    background-color: var(--green-4) !important;
+  }
+  tr > *[style*="background:#e4d4c0"],
+  tr > *[style*="background:#f2caa4"] {
+    background-color: var(--red-3) !important;
+  }
+  tr > *[style*="background:#ECECEC"],
+  tr > *[style*="background:#C0C0C0"],
+  tr > *[style*="background:#d0d0d0"],
+  tr > *[style*="background:#EFEFEF"],
+  tr > *[style*="background: rgb(204, 204, 204)"],
+  tr > *[style*="background: #CCC"] {
+    background-color: var(--gray-4) !important;
+  }
+  div[style*="background:#d9ebff"],
+  tr > *[style*="background:#e0e0e0"],
+  tr > *[style*="background:#DEDEDE"]  {
+    background-color: var(--gray-3) !important;
   }
   .HQToggle::before, .NavToggle::before, .HQToggle::after, .NavToggle::after {
     color: var(--base-color);
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -6552,17 +6552,19 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
     background-color: var(--blue-4) !important;
   }
   *[style*="background:#a0ade3"],
-  *[style*="background:#ccccff"] {
-    background-color: var(--blue-3) !important;
-  }
-  *[style*="background:#e2e4c0"],
-  *[style*="background:#d0cfa4"] {
-    background-color: var(--brown-2) !important;
-  }
+  *[style*="background:#ccccff"],
+  *[style*="background: rgb(242, 242, 255)"],
+  *[style*="background:#ccd8ff"],
   *[style*="background:#c0d8e4"] {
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Thanks for pull request, fix welcome, but I wonder if you're also willing to replace the light background on these to use the darker colors we use? This would yield best results if the text would be light against the darker colors we use.

oowekyala:
I did that, and tested it on a few languages

Capture d’écran de 2021-03-10 14-43-14

wikipedia-dark.user.css
```diff
@@ -6546,6 +6547,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
   }
+  tr > *[style*="background:#c0cfe4"],
+  tr > *[style*="background:#9999DF"] {
+    background-color: var(--blue-4) !important;
+  }
+  tr > *[style*="background:#a0ade3"],
+  tr > *[style*="background:#ccccff"] {
+    background-color: var(--blue-3) !important;
+  }
+  tr > *[style*="background:#e2e4c0"],
```
the-j0k3r, line 6558: `#e2e4c0` is on green specrum, Id'e probably go for `--green-2`

wikipedia-dark.user.css
```diff
@@ -6546,6 +6547,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
+  tr > *[style*="background:#d0cfa4"] {
+    background-color: var(--yellow-1) !important;
+  }
+  tr > *[style*="background:#c0d8e4"] {
```
the-j0k3r, line 6562: `#c0d8e4` is on blue spectrum, so maybe a `--blue-3` intead of adding a new yellow var ;)

wikipedia-dark.user.css
```diff
@@ -6546,6 +6547,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
+  tr > *[style*="background:#EFEFEF"],
+  tr > *[style*="background: rgb(204, 204, 204)"],
+  tr > *[style*="background: #CCC"] {
+    background-color: var(--gray-4) !important;
+  }
+  div[style*="background:#d9ebff"],
+  tr > *[style*="background:#e0e0e0"],
+  tr > *[style*="background:#DEDEDE"]  {
+    background-color: var(--gray-3) !important;
+  }
```
the-j0k3r, line 6592: All in all its a good PR, I would instead of specifying `tr > *` probably try to make it a more general selector, this way we could try catch as many inline styles with these rules.

That is if at all possible.

other than previous comments I think upon chnages this would be good to go @AfroThundr3007730 for thoughts.

thx again @oowekyala for this good efforts.

wikipedia-dark.user.css
oowekyala: That's right

wikipedia-dark.user.css
oowekyala: Those are used in latin conjugation tables, if we move them to green I think there's not enough difference between the different sections anymore and I find this ugly... Compare

<details><summary>Current (colorful)</summary>

Capture d’écran de 2021-03-11 09-15-48

</details>

<details><summary>With green-2 and green-1 instead of yellow/pink:</summary>


Capture d’écran de 2021-03-11 09-16-25

</details>

Also I would name these color yellows, and apparently [this website agrees](https://www.colorhexa.com/e2e4c0): they name #e2e4c0 "light grayish yellow", and #e0e0b0 is "very soft yellow".

wikipedia-dark.user.css
oowekyala: > All in all its a good PR, I would instead of specifying tr > * probably try to make it a more general selector, this way we could try catch as many inline styles with these rules.

That's true, maybe using just `*` would be ok?

the-j0k3r: If it works

wikipedia-dark.user.css
```diff
@@ -6546,6 +6546,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
+  *[style*="background:#9999DF"] {
+    background-color: var(--blue-4) !important;
+  }
+  *[style*="background:#a0ade3"],
+  *[style*="background:#ccccff"] {
+    background-color: var(--blue-3) !important;
+  }
+  *[style*="background:#e2e4c0"],
+  *[style*="background:#d0cfa4"] {
+    background-color: var(--yellow-1) !important;
```
the-j0k3r, line 6559: `#e2e4c0` as yellow-1 is way overkill youde be better sending these to brown colors.

Reserve very bright colors on bg for warning boxes borders etc

wikipedia-dark.user.css
```diff
@@ -6546,6 +6546,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
   }
+  *[style*="background:#c0cfe4"],
+  *[style*="background:#9999DF"] {
+    background-color: var(--blue-4) !important;
+  }
+  *[style*="background:#a0ade3"],
+  *[style*="background:#ccccff"] {
+    background-color: var(--blue-3) !important;
+  }
+  *[style*="background:#e2e4c0"],
```
the-j0k3r, line 6557:
```suggestion
```

wikipedia-dark.user.css
```diff
@@ -6546,6 +6546,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
+  *[style*="background:#d0cfa4"] {
```
the-j0k3r, line 6558:
```suggestion
```

wikipedia-dark.user.css
oowekyala: I think you're right, brown-2 looks good

oowekyala: brown-3 looks too dark I think.. Also I can't find a place where there is #d0cfa4 anymore so maybe this is not so important

oowekyala: How do you make github print the color with its hex?

wikipedia-dark.user.css
```diff
@@ -6546,6 +6546,49 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
+  *[style*="background:#99cc99"],
+  *[style*="background:#99CC99"] {
+    background-color: var(--green-3) !important;
+  }
+  *[style*="background:#cfedcc"] {
+    background-color: var(--green-4) !important;
+  }
+  *[style*="background:#e4d4c0"],
+  *[style*="background:#f2caa4"] {
+    background-color: var(--red-3) !important;
```
the-j0k3r, line 6577:
```suggestion
    background-color: var(--something-not-red) !important;
```

e.g. dark orange-ish `#4d2b01`

the-j0k3r:
@AfroThundr3007730 would you look it over? if so your button ;)

wikipedia-dark.user.css
AfroThundr3007730: You just type the hex code surrounded by backticks.
```
 `#3300cc`
 ```
 And you get `#3300cc`.

AfroThundr3007730:
@the-j0k3r @oowekyala Looks good to me. Thanks for working on this.

oowekyala:
Thank you both for reviewing & merging!

the-j0k3r:
Thank you for giving us the opportunity with the share ;)

PR #166, opened 2021-03-15
Fix Wiktionary backgrounds

oowekyala:
Fix some backgrounds on the wiktionary:
* the sidebar is white in the French wiktionary, also, the "see also" banner is too light
* In the german wiktionary, see eg [this page](https://de.wiktionary.org/wiki/Ungarisch), where some headers are whitish, and the translation table too (there are many shades of white in this table alone...)

Screens:
Capture d’écran de 2021-03-15 20-57-07

Capture d’écran de 2021-03-15 20-55-42

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -5829,6 +5829,7 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
     color: var(--gray-a) !important;
   }
   .bandeau-simple.bandeau-niveau-information,
+  .bandeau-section.bandeau-niveau-information,
   .bandeau-discussion.bandeau-niveau-information {
     border-color: var(--blue-4) !important;
     background-color: var(--blue-2);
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -5547,6 +5547,21 @@ regexp("^https://(foundation|donate)\.wikimedia\.org.*$") {
...
 }
+@-moz-document domain("de.wiktionary.org") {
+  div.mw-parser-output h2, div.mw-parser-output h3 {
+    background-color: transparent !important;
+  }
+  div[style*="background-color: #EFF5FB"],
+  div[style*="background-color:#EAF3FC"],
+  table[style*="background-color:#F5FAFE"],
+  td[style*="background-color: #F7F7F9"] {
+      background-color: transparent !important;
```
the-j0k3r, line 5558:
```suggestion
      background-color: var(--gray-2_75) !important;
```

That is more like it, transparent is not ideal.

the-j0k3r:
@AfroThundr3007730 your button my friend.

Issue #170, opened 2021-04-08
various issues on german wiki - bright background/missing colors


hauk92:
Sorry this issue got a bit large, feel free to yell at me to split it.
Also there are probably a lot more issues regarding election results, see https://de.wikipedia.org/wiki/Liste_von_Wahlen

* **Browser**: Firefox
* **Operating System**: Windows 10
* **Link to page with the issue**: 
### Bright background:
https://de.wikipedia.org/wiki/Wikipedia:Hauptseite
https://de.wikipedia.org/wiki/Wikipedia:Miniaturenwettbewerb
https://de.wikipedia.org/wiki/Wikipedia:Sprachen
https://de.wikipedia.org/wiki/Inuit_Ataqatigiit#Einzelnachweise
https://de.wikipedia.org/wiki/Monaco#Klima
### Missing colours etc:
https://de.wikipedia.org/wiki/Parlamentswahl_in_Gr%C3%B6nland_2021
https://de.wikipedia.org/wiki/Parlamentswahl_in_Gr%C3%B6nland_2021#Umfragen
https://de.wikipedia.org/wiki/Bundestagswahl
https://de.wikipedia.org/wiki/Bundestagswahl#Ergebnisse
https://de.wikipedia.org/wiki/Wikipedia:Statistik#Statistiken_zu_sehr_aktiven_Autoren

* **Screenshot**:
### Bright background:
![grafik](https://user-images.githubusercontent.com/47572039/114078591-72ed9580-98a9-11eb-877b-9f75379dfbde.png)
![grafik](https://user-images.githubusercontent.com/47572039/114078721-90226400-98a9-11eb-9141-bceac1307a34.png)
![grafik](https://user-images.githubusercontent.com/47572039/114078758-9c0e2600-98a9-11eb-99eb-69d130503eff.png)
![grafik](https://user-images.githubusercontent.com/47572039/114078794-a7615180-98a9-11eb-953c-deb5ae61493d.png)
![grafik](https://user-images.githubusercontent.com/47572039/114080601-dd073a00-98ab-11eb-868a-709e267eaaf4.png)
### Missing colours etc:
![grafik](https://user-images.githubusercontent.com/47572039/114079505-9107c580-98aa-11eb-8755-52fb6932464d.png)
![grafik](https://user-images.githubusercontent.com/47572039/114079340-61f15400-98aa-11eb-86cf-9194f31c0519.png)
[...]

Issue #167, opened 2021-03-24
Wikipedias result board for racing records looks pretty horrible


JesperVa:
* **Browser**: Chrome
* **Operating System**: Windows 
* **Link to page with the issue**:  https://en.wikipedia.org/wiki/2020_Formula_One_World_Championship#World_Drivers'_Championship_standings
* **Screenshot**: https://i.imgur.com/lovTtRA.png

Pretty hard to see the any of the standings and the default colours are not very dark-mode friendly.

the-j0k3r:
These are inline styles, we cant cover 100% of all cases, see https://github.com/StylishThemes/Wikipedia-Dark#limitations

It can be fixed so PR welcome.

JesperVa:
My bad just copied the issue link in the style, I'll see if I can have a look at this issue and see if these can be improved.

Might push up a PR in a few days.

PR #171, opened 2021-04-12
Various fixes

Atexor:
Hello,
As I promised on previous PR - another bunch of various fixes. There are many improvements to article translation tool, search boxes, infoboxes, borders, tables, titles, polish or german subpages (needed to remove few because I saw @AfroThundr3007730 added similarr rules for them too) and more I forgot now where.

It fixes all issues for #167 and almost all #170 (even with charts colors, images can't be fixed).

Yes, I know about `!imporant` you don't like. Many elements couldn't be couldn't be overwritten because of being overwritten by more general rules of Dark Wikipedia style or by default Style. Or needed a lot longer rules to match these fixes.

Few examples of fixes below:
s1

Have a nice day.

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -485,7 +485,7 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
     color: var(--gray-1);
   }
   h1, h2, h3, h4, h5, h6, #footer li,
-  .infobox th[style*="background"]:not([style*="#eaecf0"]),
+  .infobox th[style*="background"]:not([style*="#eaecf0"]):not([style*="background-color:#b0c4de;"]):not([style*="background-color: #f9f"]):not([style*="background-color:#b0c4de"]):not([style*="background-color:#FFD"]):not([style*="background-color: #91FAFA"]):not([style="background-color: #b0c4de"]),,
   div[style*="color"]:not([style*="black;"]):not([style*="back"]):not([style*="966"]):not([style*="red"]),
   input[type="search"], input[type="submit"], .oo-ui-buttonElement-button,
   .oo-ui-buttonElement-button:hover, input[type="number"], select,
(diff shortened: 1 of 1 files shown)


wikipedia-dark.user.css
```diff
@@ -1,7 +1,7 @@
 /* ==UserStyle==
 @name         Wikipedia Dark
 @namespace    StylishThemes
-@version      3.5.2
+@version      3.5.3
```
AfroThundr3007730, line 4: Please don't version bump in a PR. That will happen as part of the release process.

wikipedia-dark.user.css
```diff
@@ -645,6 +645,176 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
...
+  }
+  .uls-settings-trigger {
+    border-color: var(--gray-2) !important;
+  }
+  .wikibase-toolbar-container.wikibase-toolbar-container {
+    background: transparent !important;
+  }
+  .mw-parser-output .maf-info,
+  .mw-parser-output .wwikipedii tr {
+    background: var(--gray-18);
```
AfroThundr3007730, line 686: Just a nit really, but with a few exceptions, `background` should be replaced with `background-color` as a more specific property to modify color without affecting other `background-*` properties that may be in place. We do the ssme thing for borders.

Exceptions to this would be when you intentionally want to overwrite the full background property (e.g. replace `background-image`).

wikipedia-dark.user.css
```diff
@@ -645,6 +645,176 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
...
+  .cx-translationlist__header {
+    color: var(--white);
+  }
+  .cx-translationlist,
+  .cx-dashboard-sidebar__help,
+  .cx-translator {
+    box-shadow: 0 1px 1px var(--gray-5);
+  }
+  .tux-message-editor .messagekey .caret {
+    border-top: 4px solid var(--gray-c);
```
AfroThundr3007730, line 801: Further on the `border` property, if the only thing you're changing is the color (which should be the case 99% of the time), you should use just `border-color` here, or when necessary, one of the `border-*-color` properties.

So in this case, if you're only modifying color, it would be:
```
border-top-color: var(--gray-c);
```

AfroThundr3007730 (changes requested):
Some pretty gnarly selectors here, and a lot of `!important`, but the potential collateral damage should be minimal. Looking good so far, just a few things to be tweaked before this is merged (mentioned in the comments above).

Also as a note, either the PR or commits should reference the issue(s) they are involved with. Ex: `Various style fixes (fixes #167, fixes #170)`. I'll add this info before merging anyway.

wikipedia-dark.user.css
Atexor: Allright, didn't know about that.

wikipedia-dark.user.css
Atexor: Yeah, I know about it but I noticed very many `background` rules in style, where background-color should be enough. All in all, it's better not to duplicate potential mistakes . Thx.

wikipedia-dark.user.css
Atexor: Nice catch, forgot about it.

Atexor:
Allright, fixed. If you want, in the future I can try to match elements with a lot longer selectors instead using `imporant` if possible (if it's possible). I tried to find a compromise between `important` rule and path length.

Atexor (commented):
Closing review

AfroThundr3007730 (approved):

AfroThundr3007730:
Looks good. Thanks again for the PR.

PR #173, opened 2021-04-17
Fix survey background color

TotalCaesar659:
Sometimes there are goes survey: "Does this page load fast enough?" "Yes/Not sure/No". For more information, see this survey's [privacy statement](https://foundation.wikimedia.org/wiki/Performance_Perception_Survey_Privacy_Statement).

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -4313,6 +4313,10 @@ regexp("https?:\/\/([\w\-]{2,}\.m\.(wiki(pedia|books|news|quote|source|versity|v
     filter: invert(0);
   }
   /* Background fixes */
+  /* survey */
+  .ext-quick-survey-panel, .ext-qs-loader-bar {
+    background-color: var(--gray-2);
+  }
   .minerva--amc-enabled.action-history .mw-history-compareselectedversions,
   .ns-special .pre-content, .ns-special #content form {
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Thanks for PR.

2 comments.

1) if this is in ru site only why use a general section?
2) why add a new rule instead of adding the selector to an existing rule with same property

TotalCaesar659:
1. I have a doubt that this survey is only for RU zone.
2. I don't know where would be ф proper proper place for this, so you can tell me. And separate section would be faster to find if there are any problems.

TotalCaesar659:
> 1. I have a doubt that this survey is only for RU zone.
> 2. I don't know where would be a proper proper place for this, so you can tell me. And separate section would be faster to find if there are any problems.

the-j0k3r:
We dont support blogs on this style. Just FYI

TotalCaesar659:
I didn't ask support for this blog and don't understand why you thought about it. I gave that URL just for proof that the survey is not for RU zone - there is a picture under "_The micro survey simply asks users on Wikipedia articles, in their own language, if they think that the current page loaded fast enough:_"

the-j0k3r:
> I didn't ask support for this blog and don't understand why you thought about it

Cause no other links pop this up for me, have never seen it

PR #180, opened 2021-05-28
Removal of problematic code

ColonelGerdauf:
A segment of code is prohibiting custom colours from showing, which is problematic for when one wishes to see colour previews from the wiki's pages about colours. After reviewing the code segment in question, it appears like a cobbled-together mess that does not offer a lot of room for dynamic situations. If you wish to add back the sub-table grey, then I implore you to reconsider the coding logic before doing so.

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -1314,26 +1314,6 @@ regexp("https?:\/\/wiki\.(archlinux|mozilla)\.(org|jp)\/.*$") {
     border-top: 1px solid var(--gray-4);
     color: var(--gray-c);
   }
-  table.wikitable > tr > th,
-  table.wikitable > * > tr > th:not([style*="#ABC;"]):not([style*="#fdb"]):not([style*="#fed"]):not([style*="#fef"]):not([style*="#d4f"]):not([style*="0; color:#000000"]):not([style*="FF; color:#000000"]),
-  td[style*="background: #ececec;" i], th[style*="background:#eee" i],
-  th[style*="background-color: #eee" i], tr[style*="background: #dddddd" i],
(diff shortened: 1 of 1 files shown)


the-j0k3r (changes requested):
Cannot remove this willy nilly, it will cause all unstyled colors that havent yet been replaced with dark alternatives to open a bug report floodgates.

I suggest you report your issue and get it fixed or fix it by not removing this, its OK to add to it.

In an ideal world all unknown colors would be already replaced but its impossible so until a better solution comes along we need ugliness like this because inline color styles are utter BS and thats how wikis operate, blame it on whoever inserts inline colors styles instead of universal CSS classes.

TLDR see https://github.com/StylishThemes/Wikipedia-Dark#limitations

ColonelGerdauf:
Easier said than done, I am afraid. Like I said, the part of code that I had removed was so elongated and hacky that there was no way for me to write an additional code condition to allow for the colours to show; it offered zero room for doing so.

the-j0k3r:
> Easier said than done, I am afraid. Like I said, the part of code that I had removed was so elongated and hacky that there was no way for me to write an additional code condition to allow for the colours to show; it offered zero room for doing so.

Its easier done than what you may think, but since you wont report the actual issue, we wont ever know.

I agree this block looks insane and it is, but, this is the only way to keep going forward until such time when all inline colors in all pages is substituted by the dark counter parts.

So no.

ColonelGerdauf:
I have just reported the issue a while ago, if you care to look

the-j0k3r:
>it offered zero room for doing so.

It offers plenty room, it add to the insanity but theres no choice if you want a dark style for insane sites like this that use inline color styles.
check out the fix.

Issue #181, opened 2021-06-01
Colour boxes are not showing where intended


ColonelGerdauf:
* **Browser**: Firefox
* **Operating System**: Windows 10
* **Link to page with the issue**: https://en.wikipedia.org/wiki/Red or any other page about colours
* **Screenshot**:
What it should have looked like ideally:
![ideal](https://user-images.githubusercontent.com/6895655/120258746-9e18b380-c260-11eb-8c5e-691f8d267390.png)

And what it looks like under the current code:
![bugged](https://user-images.githubusercontent.com/6895655/120258821-ba1c5500-c260-11eb-9919-4efca27252bb.png)

ColonelGerdauf:
I should repeat from my pull request that the code block responsible for causing this discrepancy is a disorganized mess, making it impossible for me to figure out a way to work out this issue without removing the part of code altogether.

the-j0k3r:
Easy fix add `:not([style*="000"])` to the .infobox line, should clear up most of these if not all.

@AfroThundr3007730

PR #182, opened 2021-06-01
Patch 1

ColonelGerdauf:
Used a suggestion and extended it to boxed sub-headers where colours are supposed to display. Now that I am aware that there is indeed a simple solution to my original problem, I would like to have the fix extended to another location which could benefit from such solution.


the-j0k3r:
#181 is already fixed and fix committed.

ColonelGerdauf:
Thank you, but I have added an additional area, that appears to also fix things for another section that was not previously displaying as intended

Sorry about that, The original commit was that same one you noted issues with, the next one brought your 181 fix, and the last was my own tweak

the-j0k3r:
> Sorry about that, The original commit was that same one you noted issues with, the next one brought your 181 fix, and the last was my own tweak

This wont work, because the fix for #181 you added is conflicting with the existing and thus causes conflicts.

Capture

also such tweak you added all though it looks harmless please link URL and explain what it fixes.

You will need to start with a clean PR though as I dont have access to a dev environment, only via web on a clunky interface.

Ill close this sorry, and hope a clean PR with explanation comes along.

Thx

ColonelGerdauf:
Thank you. I am on it

PR #190, opened 2021-11-03
Major Improvements

blyad2137:
- added support for sonichu wiki, yes it was so simple (#114)
- added support for diff.mediawiki
- added customizable scrollbar
- fixes at mediawiki
- changed multiple white backgrounds to dark
- inverted dozen of images to white/appropriate color
- in some ways improved code to do not overwrite other elements that shouldn't be changed
- some minor fixes

more to come soon

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -1,7 +1,7 @@
 /* ==UserStyle==
 @name         Wikipedia Dark
 @namespace    StylishThemes
-@version      3.5.10
+@version      3.5.11
 @description  Wikipedia Dark theme
 @author       StylishThemes
 @homepageURL  https://github.com/StylishThemes/Wikipedia-Dark
@@ -40,12 +40,40 @@
   Fixed "Fixed" <<<EOT
   fixed EOT;
 }
+@advanced dropdown custom_scrollbar "Custom Scrollbar:" {
+		1. "*ON" <<<EOT ::-webkit-scrollbar-track {
+             -webkit-box-shadow: inset 0 0 0px rgba(0, 0, 0, 0.3);
+             border-radius: 0px;
+             background-color: #212121;
+        }
+        html[hide-scrollbar="true"] ::-webkit-scrollbar{display:none}
+        ::-webkit-scrollbar {
+             width: var(--custom_scrollbar_width);
+             background-color: #212121;
+        }
+        ::-webkit-scrollbar-thumb {
+             border-radius: 0px;
+             -webkit-box-shadow: inset 0 0 6px rgba(0, 0, 0, .6);
+             background-color: var(--custom_scrollbar_color);
+        }
+		::-webkit-scrollbar-corner {
+			 background-color: #212121;
+			 border-color: #212121;
+		}EOT;
+	2. "OFF" <<<EOT EOT;
+}
+@advanced text custom_scrollbar_width "Width of custom scrollbar:(px, em, rem, vh etc.)" 1em
+@advanced text custom_scrollbar_color "Color of custom scrollbar:(hex, rgb, hsla or variables)" var(--assess-1)
+
+
 @preprocessor uso
 ==/UserStyle== */
 @-moz-document
-regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
+regexp("https?:\/\/.*wik(i|t).*(org|jp).*"), domain("sonichu.com"), domain("diff.wikimedia.org") {
   /* Global color palette */
   :root {
+	  --custom_scrollbar_width: /*[[custom_scrollbar_width]]*/;
+	  --custom_scrollbar_color: /*[[custom_scrollbar_color]]*/;
     --base-color: /*[[base-color]]*/;
     --base-color-rgb: /*[[base-color-rgb]]*/;
     --base-color-rgba-3: rgba(var(--base-color-rgb), 3);
@@ -170,6 +198,7 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
     --cm-14: #aeaeae;
     --cm-15: #ad9361;
   }
+	/*[[custom_scrollbar]]*/
   .cm-comment, .cm-link {
     font-style: italic;
   }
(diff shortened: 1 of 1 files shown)


the-j0k3r:
Thank you very much for reaching out with this contribution.

Some of it is not OK. needs reverting/not adding at all.

Will look around.

Thanks.

wikipedia-dark.user.css
```diff
@@ -1,7 +1,7 @@
 /* ==UserStyle==
 @name         Wikipedia Dark
 @namespace    StylishThemes
-@version      3.5.10
+@version      3.5.11
```
the-j0k3r, line 4: Please don not include version bumps, this is strictly 100% for internal use.

wikipedia-dark.user.css
```diff
@@ -40,12 +40,40 @@
...
+		::-webkit-scrollbar-corner {
+			 background-color: #212121;
+			 border-color: #212121;
+		}EOT;
+	2. "OFF" <<<EOT EOT;
+}
+@advanced text custom_scrollbar_width "Width of custom scrollbar:(px, em, rem, vh etc.)" 1em
+@advanced text custom_scrollbar_color "Color of custom scrollbar:(hex, rgb, hsla or variables)" var(--assess-1)
+
+
```
the-j0k3r, line 68: Scroll bar code does not belong here see https://github.com/StylishThemes/Overlay-Scrollbars

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	.other-project-link div, .central-featured-lang small, .central-featured-lang strong {
+		filter: brightness(1);
+	}
+	.footer-sidebar-icon.sprite.svg-wikipedia_app_tile {
+		filter: invert(.835) hue-rotate(180deg);
+	}
+	span.jsl10n.sprite.svg-badge_ios_app_store, span.jsl10n.sprite.svg-badge_google_play_store, span.jsl10n.sprite.svg-badge_google_play_store:hover {
+		color: transparent;
+	}
+	[src="//upload.wikimedia.org/wikipedia/donate/a/ac/Close_oojs.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/a/a1/OOjs_UI_icon_caretDown.svg/20px-OOjs_UI_icon_caretDown.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/d/d8/Stylus-logo.svg/120px-Stylus-logo.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/4/4a/Dragon-149393.svg/20px-Dragon-149393.svg.png"] {
```
the-j0k3r, line 441: Please use short form

`[src="//upload.wikimedia.org/wikipedia/donate/a/ac/Close_oojs.png"]`

to

`[src*="Close_oojs.png"]`


and so on.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	}
+	.vector-user-menu-legacy #pt-anonuserpage {
+		color: #a4a9ad;
+	}
+	.vector-user-menu-legacy #pt-anonuserpage, .vector-user-menu-legacy #pt-userpage a {
+		background-image: none;
+		position: relative;
+	}
+	.vector-user-menu-legacy #pt-anonuserpage::before, .vector-user-menu-legacy #pt-userpage a::before {
+		content: url('https://pl.wikipedia.org/w/skins/Vector/resources/skins.vector.styles.legacy/images/user-avatar.svg?b7f58');
```
the-j0k3r, line 461: Please include base64 images, linking to URL is not idea.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+		filter: invert(.835);
+	}
+	th.infobox-above[style="background: #99c2ff;"] {
+		background-color: var(--blue-4) !important;
+	}
+	th.infobox-above.summary[style="font-style: italic; background: #CCCCFF; padding: 0.25em 1em; font-size: 125%;"] {
+		background-color: var(--purple-2) !important;
+		color: var(--white) !important;
+	}
+	table.wikitable.plainrowheaders.wikiepisodetable tbody tr th span[style="background-color:white;padding:1px;display:inline-block;line-height:50%"] {
```
the-j0k3r, line 483: for specificity try alway the short values

`span[style="background-color:white;padding:1px;display:inline-block;line-height:50%"]`

this should be `[style="background-color:white"]` also use this in other areas this is the way to replace colors to match only what is needed.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+		background-color: var(--grey-1) !important;
+	}
+	.rt-settingsLink {
+		background-image: none;
+		position: relative;
+	}
+	.rt-settingsLink::before {
+		content: url('data:image/svg+xml,%3C%3Fxml%20version%3D%221.0%22%20encoding%3D%22utf-8%22%3F%3E%0D%0A%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20viewBox%3D%220%200%2024%2024%22%3E%0D%0A%20%20%20%20%3Cpath%20fill%3D%22%23555%22%20d%3D%22M20%2014.5v-2.9l-1.8-.3c-.1-.4-.3-.8-.6-1.4l1.1-1.5-2.1-2.1-1.5%201.1c-.5-.3-1-.5-1.4-.6L13.5%205h-2.9l-.3%201.8c-.5.1-.9.3-1.4.6L7.4%206.3%205.3%208.4l1%201.5c-.3.5-.4.9-.6%201.4l-1.7.2v2.9l1.8.3c.1.5.3.9.6%201.4l-1%201.5%202.1%202.1%201.5-1c.4.2.9.4%201.4.6l.3%201.8h3l.3-1.8c.5-.1.9-.3%201.4-.6l1.5%201.1%202.1-2.1-1.1-1.5c.3-.5.5-1%20.6-1.4l1.5-.3zM12%2016c-1.7%200-3-1.3-3-3s1.3-3%203-3%203%201.3%203%203-1.3%203-3%203z%22%2F%3E%0D%0A%3C%2Fsvg%3E');
```
the-j0k3r, line 491: bas64 image

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	}
+	.mw-parser-output a[href$=".pdf"].external::before {
+		content: url('https://upload.wikimedia.org/wikipedia/commons/2/23/Icons-mini-file_acrobat.gif');
+		position: absolute;
+		top: 0;
+		right: 0 !important;
+		margin: 0;
+		filter: invert(.835) hue-rotate(180deg) saturate(3);
+	}*/
+	[src="//upload.wikimedia.org/wikipedia/commons/thumb/8/80/Wikipedia_laurier_W.png/67px-Wikipedia_laurier_W.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/b/b4/Video_Camera_Icon.svg/15px-Video_Camera_Icon.svg.png"] {
```
the-j0k3r, line 514: https://github.com/StylishThemes/Wikipedia-Dark/pull/190#discussion_r742563886

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+		filter: invert(1) hue-rotate(180deg);
+	}
+	td.MainPageBG[style="width:55%;border:1px solid #cef2e0;background: #f5fffa;vertical-align:top;color:#000;"], td.MainPageBG[style="width:55%; border:1px solid #cef2e0; background:#f5fffa; vertical-align:top; color:#000;"] {
```
the-j0k3r, line 517: https://github.com/StylishThemes/Wikipedia-Dark/pull/190#discussion_r742563886

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	[src="//upload.wikimedia.org/wikipedia/commons/thumb/6/6f/Octicons-terminal.svg/24px-Octicons-terminal.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/8/88/At_sign.svg/15px-At_sign.svg.png"], [src="/static/images/mobile/copyright/wikipedia-wordmark-fr.svg"], [src="/static/images/mobile/copyright/wikipedia-tagline-pt.svg"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/b/bd/Wikipedia-logo-v2-bw.svg/100px-Wikipedia-logo-v2-bw.svg.png"], .mw-ui-icon-wikimedia-language:before {
+		filter: invert(1);
+	}
+	#p-lang .mw-interlanguage-selector, #p-lang .mw-interlanguage-selector:active {
+		background-image: none;
+		position: relative;
+	}
+	#p-lang .mw-interlanguage-selector::before, #p-lang .mw-interlanguage-selector:active::before {
+		content: '';
+		background-image: url('https://en.wikipedia.org/w/extensions/UniversalLanguageSelector/resources/images/language-base20.svg?b7954');
```
the-j0k3r, line 594: base64 image

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+		border-color: var(--gray-69) !important;
+	}
+	.mw-highlight .nv {
+		color: var(--blue-1);
+	}
+	[style="background-color: #DFD;border: 1px solid #6B6; font-size: 85%;padding: 5px 8px;border-radius:10px;"] {
+		border-color: var(--green) !important;
+		background-color: var(--green-2) !important;
+	}
+	table.mw-collapsible.mw-collapsed.mw-made-collapsible[style="width:100%;margin:0;padding:0;border:1px solid #999999;border-collapse:collapse;background:#FFFFFF"], table.mw-collapsible.mw-made-collapsible[style="width:100%;margin:0;padding:0;border:1px solid #999999;border-collapse:collapse;background:#FFFFFF"] {
```
the-j0k3r, line 695: again here and everywhere, this is overill to match the whole inline style.

In all such areas ONLY use the bits you need, and this WAY too heavy selectors is 100% going to break something else.

cannot accept this at all

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+		background-color: var(--gray-2) !important;
+		border-color: var(--gray-69) !important;
+	}
+	div[style="font-weight:bold;background-color:#CCCCFF;"] {
```
the-j0k3r, line 699: We already have areas with colorsm you should not add rules when you could add the selector to already existing rule.

also its wrong this should be all you need `[style*="background-color:#CCCCFF;"]`

This is the way its done.

wikipedia-dark.user.css
```diff
@@ -767,7 +1082,6 @@ regexp("https?:\/\/wikiless\.org\/.*$") {
   .infobox tr[style*="background:#F6E5A8;"] th, td[style*="background:#AED0F0"],
   #mw-content-text .infobox-above[style*="background-color: lightblue"],
   #mwAg .infobox-above[style*="background-color: lightblue"],
-  #mw-content-text .infobox-above[style*="background: #CCCCFF;"],
```
the-j0k3r, line 770: do not remove anything thank you

wikipedia-dark.user.css
```diff
@@ -1152,7 +1466,6 @@ regexp("https?:\/\/wikiless\.org\/.*$") {
...
...-maintable):not(.floatright):not([style="width:100%; border:0px; border-collapse:collapse"]):not([style*="#f1f5"]):not([style*="#CED"]):not([style*="background:none;"]):not([style*="background:none;"]):not([cellspacing="0"]):not([style="text-align:center; width:100%"]):not([style*="E6F"]):not([style*="ffff"]):not([style*="E9F"]):not([style*="fff5"]):not([style*="E9E"]):not(.authortemplate):not(.fmbox-editnotice):not([style*="Alice"]):not([style*="efe"]):not([style*="transparent"]):not(.tmbox):not(.standard-talk):not(.cmbox-notice):not([style*="e4f2e4"]):not([style*="feeeee"]):not([style*="dfd"]):not(.processheadertemplate):not(.statstable):not([style*="eeeeff"]):not([style*="yellow"]):not([style*="#f7f8ff"]):not(.mw-alerte):not(.mp_header):not([style*="#FFDBDB"]):not([style*="#F0F8FF"]):not([style*="background: transparent"]):not([style*="background-color:transparent"]):not([style*="rgba(0,0,255,0.1)"]):not(.vertical-navbox):not([style*="#F5FAFF"]):not([style*="background color:transparent"]):not(#cite-err-report):not(.tmbox-notice):not([style*="background:transparent"]):not([style*="f5faff"]):not(.messagebox):not(#sisterwikis):not([class*="Caution"]):not(.tmbox-style):not(.ibox2),
```
the-j0k3r, line 1155: NEVER ever ever will this be merged.

Its insane but because of all inline styles its needed, just because you working on one page, this style supports many many sites and many inline styles, what you did here will break a ton of work.

Nope thanks.

wikipedia-dark.user.css
```diff
@@ -3974,7 +4287,6 @@ regexp("https?:\/\/wikiless\.org\/.*$") {
   img[src*="Portal.svg"], img[src*="Information.svg"],
   img[src*="Global_thinking.svg"], img[src*="Clipboard.svg"],
   img[src*="Darkgreen_flag_waving.svg"], img[src*="Stock_post_message.svg"],
-  img[src*="Featured_article_star_-_cross.svg"],
```
the-j0k3r, line 3977: do not remove anything like this, it exists for a reason.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	}
+	[src="//upload.wikimedia.org/wikipedia/commons/thumb/6/66/IPv4_address_structure_and_writing_systems-en.svg/300px-IPv4_address_structure_and_writing_systems-en.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/1/15/Ipv6_address.svg/300px-Ipv6_address.svg.png"] {
+		filter: invert(1);
+		background-color: transparent;
+	}
+	span.citation-needed-content[style="padding-left:0.1em; padding-right:0.1em; color:#595959; border:1px solid #DDD;"] {
+		color: var(--gray-a) !important;
+		border-color: var(--gray-69) !important;
+	}
+	[src="//upload.wikimedia.org/wikipedia/commons/thumb/6/6f/Octicons-terminal.svg/24px-Octicons-terminal.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/8/88/At_sign.svg/15px-At_sign.svg.png"], [src="/static/images/mobile/copyright/wikipedia-wordmark-fr.svg"], [src="/static/images/mobile/copyright/wikipedia-tagline-pt.svg"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/b/bd/Wikipedia-logo-v2-bw.svg/100px-Wikipedia-logo-v2-bw.svg.png"], .mw-ui-icon-wikimedia-language:before {
```
the-j0k3r, line 585: NO [src="//upload.wikimedia.org/wikipedia/commons/thumb/6/6f/Octicons-terminal.svg/24px-Octicons-terminal.svg.png"]

YES [src="Octicons-terminal"]

wikipedia-dark.user.css
```diff
@@ -40,12 +40,40 @@
...
+}
+@advanced text custom_scrollbar_width "Width of custom scrollbar:(px, em, rem, vh etc.)" 1em
+@advanced text custom_scrollbar_color "Color of custom scrollbar:(hex, rgb, hsla or variables)" var(--assess-1)
+
+
 @preprocessor uso
 ==/UserStyle== */
 @-moz-document
-regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
+regexp("https?:\/\/.*wik(i|t).*(org|jp).*"), domain("sonichu.com"), domain("diff.wikimedia.org") {
```
the-j0k3r, line 72: @AfroThundr3007730 have a looksee at this maybe a way to improve?

wikipedia-dark.user.css
```diff
@@ -40,12 +40,40 @@
...
   /* Global color palette */
   :root {
+	  --custom_scrollbar_width: /*[[custom_scrollbar_width]]*/;
+	  --custom_scrollbar_color: /*[[custom_scrollbar_color]]*/;
```
the-j0k3r, line 76: no scrollbars in style, thanks, use overlay scrollbars style.

wikipedia-dark.user.css
```diff
@@ -170,6 +198,7 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
     --cm-14: #aeaeae;
     --cm-15: #ad9361;
   }
+	/*[[custom_scrollbar]]*/
   .cm-comment, .cm-link {
```
the-j0k3r, line 202: no scrollbars in style, thanks, use overlay scrollbars style.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
   .ve-ui-mwAceEditorWidget .ace_editor {
     border-color: var(--cm-6);
   }
+	/*Blyad's Fixes*/
+	.central-featured-lang :hover, .central-featured-lang :hover strong::before, .other-project-link:hover, .central-featured-lang :hover small::before, .central-featured-lang :hover bdi::before, .central-featured-lang :hover span::before {
+		background-color: #0000;
```
the-j0k3r, line 418: use existing css vars

also creating far too many rules, we already have rules enough for most cases. Use those.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	.central-featured-lang :hover::before, .other-project-link:hover::before {
+		content: '';
+		position: absolute;
+		width: 100%;
+		height: 100%;
+		left: 0;
+		top: 0;
+		margin: 0;
+		background-color: var(--blue-3);
+		opacity: .54;
```
the-j0k3r, line 430: Why is it needed to override all these properties?

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	span.jsl10n.sprite.svg-badge_ios_app_store, span.jsl10n.sprite.svg-badge_google_play_store, span.jsl10n.sprite.svg-badge_google_play_store:hover {
+		color: transparent;
+	}
+	[src="//upload.wikimedia.org/wikipedia/donate/a/ac/Close_oojs.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/a/a1/OOjs_UI_icon_caretDown.svg/20px-OOjs_UI_icon_caretDown.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/d/d8/Stylus-logo.svg/120px-Stylus-logo.svg.png"], [src="//upload.wikimedia.org/wikipedia/commons/thumb/4/4a/Dragon-149393.svg/20px-Dragon-149393.svg.png"] {
+		filter: invert(1);
+	}
+	[style="color:#0645ad; font-weight:bold;"] {
+		color: var(--assess-2) !important;
+	}
+	[style="font:3.6em/1.5 Gill Sans MT, GillSansSA, Calibri, sans-serif; color:#4a4a4a; padding:5px 0; -moz-box-shadow: 0px 5px 5px -6px #ccc; -webkit-box-shadow: 0px 5px 5px -6px #ccc; box-shadow: 0px 5px 5px -6px #ccc; vertical-align:middle;"] {
```
the-j0k3r, line 447: Please dont use these insane complete as is inline styles.

wikipedia-dark.user.css
```diff
@@ -384,11 +413,297 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
...
+	.wb-langlinks-link > a:link, .wb-langlinks-link > a:visited {
+		color: #a4a9ad !important;
+	}
+	input[type="radio"][value] {
+		filter: invert(.835);
+	}
+	th.infobox-above[style="background: #99c2ff;"] {
+		background-color: var(--blue-4) !important;
+	}
+	th.infobox-above.summary[style="font-style: italic; background: #CCCCFF; padding: 0.25em 1em; font-size: 125%;"] {
```
the-j0k3r, line 479: color  `CCCCFF` already been overwritten , please condense all such overrides where the main one already exists.

wikipedia-dark.user.css
the-j0k3r: and all the other code in here is just, not up to scratch, much needs improving/condensing and generally study the existing style and try to emulate the way overrides are done.

wikipedia-dark.user.css
blyad2137: k

wikipedia-dark.user.css
blyad2137: k

wikipedia-dark.user.css
blyad2137: Short form in src can overwrite some other unwanted stuff, f.e [here](https://github.com/StylishThemes/Wikipedia-Dark/pull/190#pullrequestreview-797385979) it overwrited the USSR star(or what the hell is this)

wikipedia-dark.user.css
blyad2137: Why? This isn't going to vanish and base64 takes more space which can result in slower page loading, if you could explain this, that would be nice.

wikipedia-dark.user.css
blyad2137: Shouldn't there be `*` before equal sign? Saw this in your work but I wasn't sure how exactly it works. 
This will be changed.

wikipedia-dark.user.css
blyad2137: k

wikipedia-dark.user.css
blyad2137: okay

wikipedia-dark.user.css
blyad2137: It was overriding that USSR star, thats why I removed it.

wikipedia-dark.user.css
blyad2137: yup I got it

wikipedia-dark.user.css
blyad2137: You dont have transparent rule, I could use `transparent` value, but this also works.

wikipedia-dark.user.css
blyad2137: Nothing here is overrided, it is pseudo-element to allow me to make a nice transparent background without adding new variable,

wikipedia-dark.user.css
blyad2137: This will be changed

wikipedia-dark.user.css
blyad2137: Will do.

wikipedia-dark.user.css
the-j0k3r: yes I understand, the correct way is to add a new entry that matches more precisely the entry to fix. Because the override above was too generic you can aslo do this...

`img[src*="Featured_article_star_-_cross.svg"]:not([src*="USSR"]` for instance...

The reason why this is preferred even though it maybe a little crazy to do, is because we support many many many different wikis both domains and languages which you may not be automatically aware and some overrides in main section are for the most, not the few.

The alternative is, to add specific overrides for a domain under its own section, for instance russian sites we have

@-moz-document domain("ru.wiktionary.org") and domain("ru.wikipedia.org"), domain("ru.wiktionary.org").

so say this was a ru site you would create an entry that matches the USSR cross and turn filter off.

e.g. 

``` css
img[src*="Featured_article_star_-_cross_USSR.svg"] {
  filter: none !important;
}
```

wikipedia-dark.user.css
the-j0k3r: We do.

one here

https://github.com/StylishThemes/Wikipedia-Dark/blob/ef5f2266d594e9c63612cf76ccae3306582d2fb4/wikipedia-dark.user.css#L439-L442

and above that there is one with !important on it.

wikipedia-dark.user.css
the-j0k3r: thx, we already have crazy stuff going on here, the less additions to the current insanity, the better and also easier to maintain.

wikipedia-dark.user.css
the-j0k3r: try to look for the default color without style applied if the color matches `#99c2ff` or is the closest match, but say it doesnt exist.

OK so you are using here `background-color: var(--blue-4) !important;` but we dont have one that matches both !imporatnt and color.

The best thing would be to place it under a existing rule

https://github.com/StylishThemes/Wikipedia-Dark/blob/ef5f2266d594e9c63612cf76ccae3306582d2fb4/wikipedia-dark.user.css#L928

for `background: #CCCCFF`

https://github.com/StylishThemes/Wikipedia-Dark/blob/ef5f2266d594e9c63612cf76ccae3306582d2fb4/wikipedia-dark.user.css#L2640

perfect match

follow that logic and its will end up a cleaner result that will also match our color overrides.

Hope it helps, if not ask =)

the-j0k3r: FYI I search at https://github.com/StylishThemes/Wikipedia-Dark/blob/master/wikipedia-dark.user.css so I can work side by side adding new stuff in the editor on the other side of the screen.

Hope that helps, or your preferred approach would work also.

blyad2137:
Now it should be better (i hope)

wikipedia-dark.user.css
```diff
@@ -2099,8 +2299,8 @@ regexp("https?:\/\/wikiless\.org\/.*$") {
     background-color: var(--gray-2);
   }
   .flow-topic-titlebar, .fieldsetlike .legendtextlike,
-  .oo-ui-tagMultiselectWidget.oo-ui-widget-enabled.oo-ui-tagMultiselectWidget-outlined .oo-ui-tagItemWidget.oo-ui-widget-enabled {
-    background: var(--gray-2);
+  .oo-ui-tagMultiselectWidget.oo-ui-widget-enabled.oo-ui-tagMultiselectWidget-outlined .oo-ui-tagItemWidget.oo-ui-widget-enabled, .thumbimage {
```
the-j0k3r, line 2302: not sure if thumbimage bg can be anything else other than white, some pages have dark images and then cannot be seen

e..g https://github.com/StylishThemes/Wikipedia-Dark/blob/7636ec668496705476a429264586606d217a72e7/wikipedia-dark.user.css#L4617

blyad2137: I have changed those inline styles now and it looks like this: 
image
Pretty neat if you ask me

Its just a matter of changing dropdown menus, the search icon do not appear because its overrided by original code and you said to do not remove anything. Tho I could fix this in pseudo-element rather easly.

blyad2137: It didn't even required pseudo element, only added this `form[role=search] input[type=submit]` to transparent background

the-j0k3r: will review tomorrow, now its time for zzz

blyad2137: And about `.thumbimage`, most of the images are pure black (#000) not #222 which makes them rather visible, but I will leave this up to you.

wikipedia-dark.user.css
the-j0k3r: yes it should, if you want to know what t means read this https://www.w3schools.com/css/css_attribute_selectors.asp

wikipedia-dark.user.css
the-j0k3r: base64, will always display, its faster loading and you dont open unnecessary connections from the style to go read that file.

Obviously some optimized version of the png then converted/optimized to base64, the smaller the better.

blyad2137:
[More improvements](https://github.com/StylishThemes/Wikipedia-Dark/pull/190/commits/79bcdef822a49fceb08dd2c1616a3e9fd5d4d6bc)

- Fixed tables at meta.wikimedia
image
image

- Fixed dropdown at diff.wikimedia
image

- Inverted dozen of logos and images to white f.e:
image

- Changed dozen of info headers to dark colours f.e:
image

- Added diif.wikimedia to regexp

- Fixed few Portals:
image
image

- Changed dozen of borders to dark colours

- Some more things that i forgot

AfroThundr3007730:
@blyad2137 Are you still working on this PR? There are still several issues unresolved before this can be merged.

wikipedia-dark.user.css
the-j0k3r: IIRC this is used in an array of content and its better to leave it untouched.

See https://raw.githubusercontent.com/StylishThemes/Feature-Override-Styles/master/wikipedia-dark-img-styles.user.css which is really where special cases should be handled because it adds more flexibility per user.

However that theme needs to be updated with more specific images as they are found.

PR #193, opened 2021-11-18
wikiversity proofs

anastygnome:
Edit the theme so that  wikiversity proof do not appear as white-on-white.


the-j0k3r:
@anastygnome hello and thanks for reaching out with this contribution.

Could you post a link to the specific page please?

anastygnome:
I know it's in french but 
https://fr.wikiversity.org/wiki/Fonction_d%C3%A9riv%C3%A9e/Extremum_local

anastygnome:
after the change 
image
before the change
image

the-j0k3r:
I cant actually see what you see but looks ok, even if yellow  should be brown, not gray and we arealy have such rules so no new ones need to be created.

PR #197, opened 2022-08-07
Minor adjustments

ColonelGerdauf:
Tweaks swatch colours so that they appear as original regardless of the intrusive rule in place.
Also, a few of the cleanliness rules have been taken into account and have done the appropriate fixes.


AfroThundr3007730 (commented):
This will come at the cost of increasing the specificity of these selectors and making them harder to override later. However, after browsing WP, I haven't seen any breakage as a result of this change so far.

AfroThundr3007730 (approved):

PR #199, opened 2022-11-08
Small tweaks

ColonelGerdauf:
removed redundancies
corrected a faulty rule param
corrected group orientations [incomplete]

### DIFF
--- a/wikipedia-dark.user.css
+++ b/wikipedia-dark.user.css
@@ -228,10 +228,10 @@ regexp("https?:\/\/.*wik(i|t).*(org|jp).*") {
   .cm-mw-skipformatting {
     background-color: var(--cm-6);
   }
-  .cm-variable-2,
-  .cm-variable-3,
   span.cm-def,
   span.cm-tag,
+  .cm-variable-2,
+  .cm-variable-3,
   .cm-mw-extlink-protocol,
   .cm-mw-free-extlink-protocol,
   .cm-mw-extlink,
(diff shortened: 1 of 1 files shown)


ColonelGerdauf:
Finished the cleanup of rules. Found a few more redundancies.

When writing or modifying a rule; keep this order in mind

> objects
> conditions
> classes
> id's

the-j0k3r:
This is very well fine and dandy contribution but ignores several things.

Add version patch, this is internal development only (please dont)
No perfectionist ran on this selectors are now in separate lines (please do)

No idea how many breakages this will introduce, given the sheer insanity that we try to accommodate in a single style for a zillion pages. 

@AfroThundr3007730 for final decision.

wikipedia-dark.user.css
```diff
@@ -1490,13 +1840,19 @@ regexp("https?:\/\/wikiless\.org\/.*$") {
   .mwe-popups .mwe-popups-container .mwe-popups-thumbnail {
     background-color: var(--white) !important;
   }
-  span[style="color:black"], span[style*="color: black"],
-  span[style*="color:black"], td[style*="color: black"],
-  td[style*="color: #2C2C2C"], .mwe-popups-container a {
+  span[style="color:black"],
+  span[style*="color: black"],
+  span[style*="color:black"],
```
the-j0k3r, line 1845: one of these can be removed, span[style*="color:black"] will match span[style*="color:black"] so keep the span[style*="color:black"]

AfroThundr3007730:
From what I can see after reviewing this, you split the selectors to be one per line, then sorted them by type. It looks like there are no functional changes, aside from removing a redundant selector above.

To reduce the size of the CSS, we typically run `perfectionist` to group the selectors, which your first commit basically reverses. If we wanted to go with a one selector per line format, sorting them (as per your second commit) makes sense.

So what advantages do we gain from this expanded and sorted format? If the goal is increased legibility and maintainability, I would agree on that point. The 10% line count increase doesn't actually increase the file size much (2KB).

Still, there's not much reason to do a reformatting on its own, I think, unless this is meant to be followed up with further changes. @ColonelGerdauf and @the-j0k3r thoughts?

ColonelGerdauf:
The reason for splitting the selectors to one per line is that it makes it significantly easier to review the existing selectors and seek problematic or redundant entries.

As noted by Joker, the current code base as it was was a nightmare in terms of legacy code with immense difficulties in removing outdated entries. This now should make it easier for people to troubleshoot.

ColonelGerdauf:
Another benefit of separating the selectors to line by line is that it makes the process simple of merging selectors with the same declarations. It might make the code look a bit ugly, but in my opinion it is much easier than finding the same list of declarations written two or three times.

PR #206, opened 2023-04-27
Update wikipedia-dark.user.css

Atexor:
Hello,
Below are my fixes to new style - removing dark text, white backgrounds, titles backgrounds on infoboxes, editing articles icons or preferences stuff.

Example screen:
obraz

PR #208, opened 2023-05-28
Fixes

ghost:


AfroThundr3007730:
Rebased onto dev after merging #206

wikipedia-dark.user.css
```diff
@@ -471,10 +471,10 @@
...
   .wikitable > tr,
   .wikitable > tr > th,
-  .wikitable > tr > td,
-  .wikitable > * > tr,
+  .wikitable > tr > td:not([style*="background-color"]),
+  .wikitable > * > tr:not([bgcolor]),
   .wikitable > * > tr > th,
-  .wikitable > * > tr > td:not(.break),
+  .wikitable > * > tr > td:not(.break):not([style*="color"]):not([style*="background"]):not([bgcolor]),
   .wikitable th span[style*="background"],
```
AfroThundr3007730, line 503: This has a high potential for collateral damage, especially given the amount of custom wikitable colors and styles employed on Wikipedia. Until they have more specific fixes, this rule remains necessary.

I would be hesitant to accept this change without reviewing quite a few wikitables with color scales to ensure it has the desired effect. We man find this starts an endless game of whack-a-mole.

wikipedia-dark.user.css
```diff
@@ -419,6 +419,39 @@
...
+  .tracklist-total-length span,
+  .tracklist-total-length td {
+    background-color: var(--gray-3) !important;
+  }
+  .infobox-data {
+    background-color: var(--gray-2) !important;
+  }
+  .infobox-header {
+    border-color: var(--gray-5) !important;
+  }
```
AfroThundr3007730, line 448: These two rules no longer seem to be necessary after #206 was applied. If you find an area where it appears to be needed, we can revisit this one. Also try to avoid low specificity selectors using `!important`, where possible.

wikipedia-dark.user.css
```diff
@@ -3721,6 +3762,7 @@
   table.wikitable,
   button.formBtn {
     border: 1px solid var(--gray-5) !important;
+    background-color: var(--gray-2) !important;
   }
```
AfroThundr3007730, line 3766: This section is handling borders, so adding rules affecting backgrounds is not appropriate here (and will cause collateral damage). A more specific selector should be chosen for this rule as well.

AfroThundr3007730 (changes requested):
Many of these changes will need to be revisited after the merge of #206, which affected some of the same areas as this PR. Any changes that are still relevant would need to be moved to the most appropriate section of the stylesheet, instead of lumped in near the top.

As mentioned in the comments, tread carefully when making changes with a high potential for collateral damage, such as combining `!important` with low specificity selectors, or changing the behavior of one.

I would also suggest reviewing the [contributing guidelines](https://github.com/StylishThemes/Wikipedia-Dark/blob/master/CONTRIBUTING.md) for additional advice.

PR #211, opened 2023-06-02
Update wikipedia-dark.user.css

Atexor:
Fixed the white color background of the international Wikipedia and for several language versions when user is logged on Wikipedia (otherwise background was fine). Fixed infoboxes, navbars and weather tables fonts colors and backgrounds for Czech, German, Polish and russian versions. Changed login menu for international version and few buttons to not be white (also only while logged in).
obraz


AfroThundr3007730 (approved):
Thanks for the PR. Remember to run `npm run clean` to fix formatting issues in the future.

PR #219, opened 2024-01-18
Update wikipedia-dark.user.css

Atexor:
Fixes white sidebars, headers and navbars when user is logged in Wikipedia. Bug affected different language versions of Wikipedia.

Comparison:
obraz


ColonelGerdauf (approved):
This is very much needed; with the recent updates, the broken consistency is quite jarring.

